PR #10, opened 2017-04-20
Add USE_CUSTOM_PG_BUILD option

byucesoy:
With this commit, we add USE_CUSTOM_PG_BUILD option to be able to run tests
against custom compiled PostgreSQL build. If this environment variable evaluates
to false or it is undefined, latest available PostgreSQL release will be
installed from Debian repositories.

### DIFF
--- a/travis/install_custom_pg
+++ b/travis/install_custom_pg
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+set -eux
+
+# exit early if a custom build is not needed
+if [ -z "${USE_CUSTOM_PG}" ]; then
+  exit
+fi
+
+# clone PostgreSQL
+cd ~
+git clone -b "REL${PGVERSION//./_}_STABLE" --depth 1 git://git.postgresql.org/git/postgresql.git
+
+# we will use this to parallelize PostgreSQL compilation
+procs="$(nproc)"
+mjobs="$((procs + 1))"
+
+# configure, build and install PostgreSQL
+cd postgresql
+./configure --enable-cassert --enable-debug --with-openssl \
+    --mandir="/usr/share/postgresql/${PGVERSION}/man" \
+    --docdir="/usr/share/doc/postgresql-doc-${PGVERSION}" \
+    --sysconfdir=/etc/postgresql-common \
+    --datarootdir=/usr/share/ \
+    --datadir="/usr/share/postgresql/${PGVERSION}" \
+    --bindir="/usr/lib/postgresql/${PGVERSION}/bin" \
+    --libdir=/usr/lib/x86_64-linux-gnu/ \
+    --libexecdir=/usr/lib/postgresql/ \
+    --includedir=/usr/include/postgresql/ \
+
+make -j "${mjobs}" -s all
+make -j "${mjobs}" -s -C src/test/isolation
+sudo make install
+
+# install postgresql-common to get psql wrappers, etc.
+sudo apt-get -y -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" install postgresql-common
--- a/travis/install_pg
+++ b/travis/install_pg
@@ -4,13 +4,18 @@
 
 set -eux
 
+# exit early if a custom build is needed
+if [ -n "${USE_CUSTOM_PG}" ]; then
+    exit
+fi
+
 # always install postgresql-common
 packages="postgresql-common libedit-dev libpam0g-dev libselinux1-dev"
 
 # we set PGVERSION to 10x of the Citus version when testing Citus, so
 # only install PostgreSQL proper if it's 10 or lower
 if [ "${PGVERSION//./}" -le "100" ]; then
-  packages="$packages postgresql-$PGVERSION postgresql-server-dev-$PGVERSION"
+    packages="$packages postgresql-$PGVERSION postgresql-server-dev-$PGVERSION"
 fi
 
 # shellcheck disable=SC2086
--- a/travis/pg_travis_multi_test
+++ b/travis/pg_travis_multi_test
@@ -20,11 +20,20 @@ sudo make install
 # Change to test directory for remainder
 cd src/test/regress
 
+# if a custom build is in use, we will also run check-vanilla
+# and check-isolation along with regular regression tests
+if [ -n "${USE_CUSTOM_PG}" ]; then
+    testtargets="${testtargets} check-vanilla check-isolation"
+fi
+
 # Run tests. DBs owned by non-standard owner put socket in /tmp
 # shellcheck disable=SC2086
 make ${testtargets} || status=$?
 
 # Print diff if it exists
 if test -f regression.diffs; then cat regression.diffs; fi
+if test -f ~/postgresql/src/test/regress/regression.diffs; then
+    cat ~/postgresql/src/test/regress/regression.diffs
+fi
 
 exit $status
--- a/travis/nuke_pg
+++ b/travis/nuke_pg
@@ -7,6 +7,11 @@ set -eux
 # stop all existing instances (because of https://github.com/travis-ci/travis-cookbooks/pull/221)
 sudo service postgresql stop
 
+# remove existing installation of postgresql
+if [ -n "${USE_CUSTOM_PG}" ]; then
+    sudo apt-get --purge -y remove postgresql-*
+fi
+
 # and make sure they don't come back
 echo 'exit 0' | sudo tee /etc/init.d/postgresql
 sudo chmod a+x /etc/init.d/postgresql


travis/install_pg
```diff
@@ -4,13 +4,31 @@
...
+    packages="$packages postgresql-$PGVERSION postgresql-server-dev-$PGVERSION"
+  fi
+
+  sudo apt-get -y -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" install $packages
+else
 
-sudo apt-get -y -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" install $packages
+  # clone PostgreSQL 
+  cd ~
+  git clone -b REL9_6_2 --depth 1 git://git.postgresql.org/git/postgresql.git
```
anarazel: It strikes me as a bad idea to hardcode versions here.  I think it'd be better to:
a) just use the relevant branch, instead of a tag
b) do this for both 9.5 and 9.6.

travis/install_pg
```diff
@@ -4,13 +4,31 @@
...
+
+  # configure, build and install PostgreSQL
+  cd postgresql
+  ./configure --enable-cassert --with-openssl
```
anarazel: At least for valgrind this'll have to specify -DUSE_VALGRIND to enable full valgrind instrumentation. E.g. by doing export COPT='-DUSE_VALGRIND'.   I'm not sure whether we'd end up using the same infrastructure for that, but it's worth thinking about.

travis/install_pg
```diff
@@ -4,13 +4,31 @@
...
+  make -j8 all
```
anarazel: Hardcoding concurrency might not be the greatest approach - I suspect few of the travis instances will have that many cores available.  Maybe use `-j$(grep -c ^processor /proc/cpuinfo)`?

travis/pg_travis_multi_test
```diff
@@ -12,16 +12,31 @@ fi
...
-./configure --enable-coverage PG_CONFIG=/usr/lib/postgresql/$PGVERSION/bin/pg_config
+# If USE_CUSTOM_PG_BUILD variable is not set, we will assume custom build is
+# not required. If custom build is not required, that means PostgreSQL was
+# installed via apt, otherwise we compiled it from the source. Depending
+# on the situation, location of pg_config is different.
+if [ -z "$USE_CUSTOM_PG_BUILD" ] || ! [ $USE_CUSTOM_PG_BUILD ]; then
+  ./configure --enable-coverage PG_CONFIG=/usr/lib/postgresql/$PGVERSION/bin/pg_config
+else
+  ./configure --enable-coverage PG_CONFIG=/usr/local/pgsql/bin/pg_config
+fi
```
anarazel: Seems like it'd be better to set a $PG_CONFIG or such in the above branches, rather than repeating it later.  That way we don't need to maintain two branches here.

travis/pg_travis_multi_test
```diff
@@ -12,16 +12,31 @@ fi
...
-# Run tests. DBs owned by non-standard owner put socket in /tmp
-make ${testtargets} || status=$?
+# If USE_CUSTOM_PG_BUILD variable is not set or set to false we will only run
+# testtargets, otherwise we will run check-vanilla and check isolation along
+# with our regular regression tests.
+if [ -z "$USE_CUSTOM_PG_BUILD" ] || ! [ $USE_CUSTOM_PG_BUILD ]; then
+  make ${testtargets} || status=$?
+else
+  make check && make check-vanilla && make check-isolation || status=$?
+fi
```
anarazel: Hm, is ignoring `$testtargets` a good idea?

travis/install_pg
```diff
@@ -4,13 +4,31 @@
 
 set -eux
 
-# always install postgresql-common
-packages="postgresql-common libedit-dev libpam0g-dev libselinux1-dev"
+# If USE_CUSTOM_PG_BUILD variable is not set, we will assume custom build is
+# not required. If custom build is not required, we will install PostgreSQL
+# via apt, otherwise we will compile from the source.
+if [ -z "$USE_CUSTOM_PG_BUILD" ] || ! [ $USE_CUSTOM_PG_BUILD ]; then
```
anarazel: I'd suggest replacing this with `[ "$USE_CUSTOM_PG_BUILD" = "1" ]` or such.

byucesoy: I used `[ "${USE_CUSTOM_PG_BUILD-0}" = "0" ]`. Parameter substitution also handles if `USE_CUSTOM_PG_BUILD` is not set.

travis/install_pg
byucesoy: We will probably never use valgrind on travis and adding -DUSE_VALGRIND also requires installing valgrind package for compilation. I do not want to install a package with no benefit.

If we want to run valgrind in the future we can add it to these scripts easily.

travis/install_pg
```diff
@@ -4,13 +4,38 @@
 
 set -eux
 
-# always install postgresql-common
-packages="postgresql-common libedit-dev libpam0g-dev libselinux1-dev"
+# If USE_CUSTOM_PG_BUILD variable is not set, we will assume custom build is
+# not required. If custom build is not required, we will install PostgreSQL
+# via apt, otherwise we will compile from the source.
+if [ "${USE_CUSTOM_PG_BUILD-0}" = "0" ]; then
```
jasonmp85: Rather than adding conditionals and indenting everything, could you just add a new script (maybe `install_custom_pg`) and make each script exit early based on `USE_CUSTOM_PG_BUILD`'s value?

travis/install_pg
```diff
@@ -4,13 +4,38 @@
...
+  fi
+
+  sudo apt-get -y -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" install $packages
+else
 
-sudo apt-get -y -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" install $packages
+  # clone PostgreSQL 
+  cd ~
+  if [ "$PGVERSION" = "9.5" ]; then
+    git clone -b REL9_5_STABLE --depth 1 git://git.postgresql.org/git/postgresql.git
```
jasonmp85: Just do `git clone -b REL${PGVERSION//./_}` to replace the dot in-line. no need for conditionals.

travis/install_pg
```diff
@@ -4,13 +4,38 @@
...
+  elif [ "$PGVERSION" = "9.6" ]; then
+    git clone -b REL9_6_STABLE --depth 1 git://git.postgresql.org/git/postgresql.git
+  else
+    echo "Unsupported PostgreSQL version: $PGVERSION"
+    exit 128
+  fi
+
+  # configure, build and install PostgreSQL
+  cd postgresql
+  ./configure --enable-cassert --with-openssl
```
jasonmp85: These flags don't begin to actually emulate those used by a full PGDG build. @anarazel are we OK with that?

travis/install_pg
```diff
@@ -4,13 +4,38 @@
...
+  make -j$(grep -c ^processor /proc/cpuinfo) -s all
```
jasonmp85: This argument to make should be set at the number of cores plus one:

```bash
procs="$(nproc)"
mjobs="$(expr $procs + 1)"
make -j${mjobs} # ...
```

travis/pg_travis_multi_test
```diff
@@ -12,16 +12,33 @@ fi
...
 
-# Run tests. DBs owned by non-standard owner put socket in /tmp
-make ${testtargets} || status=$?
+# If USE_CUSTOM_PG_BUILD variable is not set or set to false we will only run
+# testtargets, otherwise we will run check-vanilla and check isolation along
+# with our regular regression tests.
+if [ "${USE_CUSTOM_PG_BUILD-0}" = "0" ]; then
+  make ${testtargets} || status=$?
+else
+  make ${testtargets} && make check-vanilla && make check-isolation || status=$?
```
jasonmp85: Don't call make three times here. In fact, it would be easiest to just check if `USER_CUSTOM_PG_BUILD` is set, then add `check-vanilla` and `check-isolation` to `testtargets`. Then you can use the existing invocation of `make` to do what you want.

travis/setup_apt
```diff
@@ -4,18 +4,23 @@
 
 set -eux
 
-# import the PostgreSQL repository key
-sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys ACCC4CF8
+# If USE_CUSTOM_PG_BUILD variable is not set, we will assume custom build is
+# not required. If custom build is not required, we will setup apt to install
+# PostgreSQL via apt, otherwise we do not need to setup anything.
+if [ "${USE_CUSTOM_PG_BUILD-0}" = "0" ]; then
```
jasonmp85: If `USER_CUSTOM_PG_BUILD` is set, just exit early.

jasonmp85:
@anarazel — Would it make sense to simply _compile_ against PGDG (non-assert) packages, then compile and test against an assert build? If we run tests against an assert build, do we really need to run them against a non-assert build, as well?

I can see compiling against a non-assert build (since it might catch a needed assert-only macro), but am not clear on the benefits of testing against a non-assert build if we're already testing against assert builds.

travis/install_pg
jasonmp85: @anarazel says to add `--enable-debug` and that's good for now.

byucesoy:
@jasonmp85 I updated the PR based on your review.

travis/pg_travis_multi_test
```diff
@@ -12,14 +12,30 @@ fi
 status=0
 testtargets="$*"
 
+# Depending on whether we compiled PostgreSQL from the source of not,
```
jasonmp85: I think you can get rid of this section.

travis/install_pg
```diff
@@ -4,6 +4,13 @@
 
 set -eux
 
+# If USE_CUSTOM_PG_BUILD variable is set, we will compile PostgreSQL from the source.
+if [ "${USE_CUSTOM_PG_BUILD-0}" != "0" ]; then
+  install_custom_pg
```
jasonmp85: Rather than having one script call the other, can you just call _both_ in the `.travis.yml` file, and then have each have the "short-circuit" logic at the top to do the right thing?

This is how `install_citus` and `install_pg` work and we used to just call both.

travis/install_custom_pg
```diff
@@ -0,0 +1,18 @@
...
+cd ~
+git clone -b REL${PGVERSION//./_}_STABLE --depth 1 git://git.postgresql.org/git/postgresql.git
+
+# we will use this to parallelize PostgreSQL compilation
+procs="$(nproc)"
+mjobs="$(expr $procs + 1)"
+
+# configure, build and install PostgreSQL
+cd postgresql
+./configure --enable-cassert --enable-debug --with-openssl
```
jasonmp85: Add `--bindir="/usr/lib/postgresql/${PGVERSION}/bin"` to the `configure` arguments so you can just have the binaries live in the same spot whether PGDG or custom.

travis/install_pg
```diff
@@ -4,6 +4,13 @@
 
 set -eux
 
+# If USE_CUSTOM_PG_BUILD variable is set, we will compile PostgreSQL from the source.
+if [ "${USE_CUSTOM_PG_BUILD-0}" != "0" ]; then
```
jasonmp85: For brevity, can you (a) abbreviate to `USE_CUSTOM_PG` and (b) just use the convention that if it's not defined or empty, it's false. Like this:

```bash
if [ -z "${USE_CUSTOM_PG} ]; then
  # ...
```

travis/setup_apt
```diff
@@ -13,13 +19,6 @@ sudo rm /etc/apt/sources.list.d/google-chrome*
 # add the PostgreSQL 9.5 repository
 sudo sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -cs)-pgdg main 9.5" >> /etc/apt/sources.list.d/postgresql.list'
 
-if [ "${PGVERSION}" == "10" ]; then
```
jasonmp85: Why was this removed?!?

travis/setup_apt
```diff
@@ -4,6 +4,12 @@
 
 set -eux
 
+# If USE_CUSTOM_PG_BUILD variable is set, we will compile PostgreSQL
```
jasonmp85: This is an optimization. I say don't modify this file at all…

jasonmp85:
This is ready to go! Merging (and HLL is even working now).

PR #16, opened 2017-05-15
Use cached PostgreSQL build

byucesoy:
We started to cache PostgreSQL build to reduce testing time. With this PR, we
modified the logic for getting PostgreSQL source such that if there is cached
PostgreSQL build, we will just pull new updates, hopefuly there are no updates
and we can skip the compiling PostgreSQL.

### DIFF
--- a/travis/install_custom_pg
+++ b/travis/install_custom_pg
@@ -7,25 +7,34 @@ if [ -z "${USE_CUSTOM_PG:-}" ]; then
   exit
 fi
 
-# in order for other PostgreSQL packages to operate correctly, we need
-# to ensure our build uses the same directories as the PGDG build; un-
-# fortunately, not all can be specified by ./configure, so we apply the
-# version-specific packaging patch used for the PGDG builds
-pkgingurl='https://anonscm.debian.org/git/pkg-postgresql/postgresql.git'
-patchurl="${pkgingurl}/plain/debian/patches/50-per-version-dirs.patch"
-
-# clone PostgreSQL
+# if there is cached PostgreSQL build, just pull new updates, hopefuly
+# there are no updates and we can skip the compiling PostgreSQL. Note
+# that travis always creates caching directories. It will be just empty
+# if there is no cache yet. We check whether the directory contains any
+# files.
 cd ~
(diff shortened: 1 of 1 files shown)


jasonmp85:
Are you certain this actually does anything? The docs say _Only modifications made to the cached directories from normal pushes are stored_, which sounds like we'd never store anything in the cache, since our postgresql building occurs on pull request builds, not push builds.

byucesoy:
Yes I am. You can check build time of [this PR](https://github.com/citusdata/citus/pull/1416) in our community repo which uses this branch. Travis tests for PR takes about 8 mins(down from 12 mins)

Pull request caches are stored but only available for that pull request. That means for every PR, we will need to compile PostgreSQL once, then for subsequent commits to that PR will use the cache.

On the other hand push caches are tied with branch's cache and they are available to all other branches(and their PRs) forked from that branch (if that branch does not have its own cache). For example when we open PR, travis will first check cache for master branch and since there is no cache for master(and never will be, because we do not build PostgreSQL for push builds), that PR will create its own cache.

It actually makes sense to build PostgreSQL for master branch once in a while, so every PR won't need to build PostgreSQL even for the first time, but this either require manual effort or something like cronjob.

travis/install_custom_pg
```diff
@@ -7,25 +7,31 @@ if [ -z "${USE_CUSTOM_PG:-}" ]; then
...
-# version-specific packaging patch used for the PGDG builds
-pkgingurl='https://anonscm.debian.org/git/pkg-postgresql/postgresql.git'
-patchurl="${pkgingurl}/plain/debian/patches/50-per-version-dirs.patch"
-
-# clone PostgreSQL
+# if there is cached PostgreSQL build, just pull new updates, hopefuly
+# there are no updates and we can skip the compiling PostgreSQL
 cd ~
-git clone -b "REL${PGVERSION//./_}_STABLE" --depth 1 git://git.postgresql.org/git/postgresql.git
+if test "$(ls -A postgresql)"; then
```
marcocitus: This will produce an error message when the directory does not exist (most of the time). Maybe use `if [ -d "postgresql" ]`.

byucesoy: Travis always creates the directory. If there is no cache, directory will be just empty. I used ls -A to ensure the directory is not empty.

PR #17, opened 2017-06-05
Changes for PostgreSQL 10

jasonmp85:
Opening to clarify what's necessary for latest PostgreSQL 10 builds.

### DIFF
--- a/travis/install_custom_pg
+++ b/travis/install_custom_pg
@@ -17,7 +17,7 @@ if [ "$(ls -A postgresql)" ]; then
   git -C postgresql pull
 else
   if [ "${PGVERSION}" -eq '10' ]; then
-    gitref='REL_10_BETA1'
+    gitref='master'
   else
     gitref="REL${PGVERSION//./_}_STABLE"
   fi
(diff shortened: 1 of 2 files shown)


travis/install_custom_pg
```diff
@@ -15,8 +15,14 @@ pkgingurl='https://anonscm.debian.org/git/pkg-postgresql/postgresql.git'
 patchurl="${pkgingurl}/plain/debian/patches/50-per-version-dirs.patch"
 
 # clone PostgreSQL
+if [ "${PGVERSION}" -eq '10' ]; then
+    gitref='REL_10_BETA1'
+else
+    gitref="REL${PGVERSION//./_}_STABLE"
+fi
+
```
anarazel: I'm not quite sure what the point of using BETA1 rather than master is.  That just seems to hide upcoming breakage?

jasonmp85: Hm. Good point. For some reason I figured features could keep going in master while the release branch diverged, but I guess the whole point is there is no release branch yet and the beta tag lives on master instead. Will just change to master.

PR #20, opened 2017-08-15
Use appropriate PostgreSQL branch

byucesoy:
PostgreSQL master branch is now stamped with 11devel and we should only
use it if we want to test against PostgreSQL 11. For PostgreSQL 9.6
tests we should use REL9_6_STABLE and for PostgreSQL 10 we should use
REL_10_STABLE (Notice extra _ after REL).

### DIFF
--- a/travis/install_custom_pg
+++ b/travis/install_custom_pg
@@ -16,10 +16,12 @@ cd ~
 if [ "$(ls -A postgresql)" ]; then
   git -C postgresql pull
 else
-  if [ "${PGVERSION}" -eq '10' ]; then
-    gitref='master'
+  if [ "${PGVERSION}" == '11' ]; then
+    gitref="master"
+  elif [ "${PGVERSION}" == '9.6' ]; then
+    gitref="REL9_6_STABLE"
   else
-    gitref="REL${PGVERSION//./_}_STABLE"
+    gitref="REL_${PGVERSION}_STABLE"
   fi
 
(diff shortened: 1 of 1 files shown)


travis/install_custom_pg
```diff
@@ -16,10 +16,12 @@ cd ~
 if [ "$(ls -A postgresql)" ]; then
   git -C postgresql pull
 else
-  if [ "${PGVERSION}" -eq '10' ]; then
-    gitref='master'
+  if [ "${PGVERSION}" == '11' ]; then
+    gitref="master"
+  elif [ "${PGVERSION}" == '9.6' ]; then
+    gitref="REL96_STABLE"
```
marcocitus: should this be `REL9_6_STABLE`?

byucesoy: Yes, I fixed it.

travis/install_custom_pg
```diff
@@ -16,10 +16,12 @@ cd ~
...
 else
-  if [ "${PGVERSION}" -eq '10' ]; then
-    gitref='master'
+  if [ "${PGVERSION}" == '11' ]; then
+    gitref="master"
+  elif [ "${PGVERSION}" == '9.6' ]; then
+    gitref="REL9_6_STABLE"
   else
-    gitref="REL${PGVERSION//./_}_STABLE"
+    gitref="REL_${PGVERSION}_STABLE"
```
jasonmp85: This breaks PostgreSQL 9.5 builds because there is no branch named `REL_9.5_STABLE`.

PR #27, opened 2017-09-11
Send regression test's output with valgrind result mail

### LINKED ISSUE #23: Valgrind tests should also check and send regression.diffs
Sometimes we may miss some tests if tests fail, but we can't see any memory problems in logs.txt. Checking regression.diffs basically means we run all regressions properly. 

https://github.com/citusdata/tools/blob/develop/valgrind/run-valgrind-tests#L53

byucesoy:
Fixes #23

### DIFF
--- a/valgrind/run-valgrind-tests
+++ b/valgrind/run-valgrind-tests
@@ -49,11 +49,19 @@ sudo chown ubuntu /usr/local/pgsql/bin/ -R
 cd src/test/regress
 make check-multi-vg VALGRIND_LOG_FILE=logs.txt || true
 
-# surprisingly this hits inbox
+attachments=""
 if [ -s logs.txt ]; then
-	mail -aFrom:valgrind-test@citusdata.com -s "[Valgrind Test Results] - Failure" -A logs.txt burak@citusdata.com metin@citusdata.com < /dev/null
+    attachments=$attachments" -A logs.txt"
+fi
+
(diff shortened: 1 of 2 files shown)


valgrind/run-valgrind-tests
```diff
@@ -51,9 +51,9 @@ make check-multi-vg VALGRIND_LOG_FILE=logs.txt || true
 
 # surprisingly this hits inbox
 if [ -s logs.txt ]; then
```
metdos: We should also check if regression.diffs has anything?

byucesoy:
@metdos do you want to have another look?

jasonmp85:
Quick question: I had an issue last month where my Heroku PG instance wasn't accepting writes, and cron fired off a bunch of mail. Did anyone get them, or did they just go locally?

metdos:
> Quick question: I had an issue last month where my Heroku PG instance wasn't accepting writes, and cron fired off a bunch of mail. Did anyone get them, or did they just go locally?

I didn't get any emails.

byucesoy:
> Quick question: I had an issue last month where my Heroku PG instance wasn't accepting writes, and cron fired off a bunch of mail. Did anyone get them, or did they just go locally?

I didn't get any emails too.

metdos:
Looks good, verified that we also send regression test's output. 🚢

PR #28, opened 2017-11-01
Add libcurl to valgrind test dependencies

metdos:
We also need to install libcurl now. Look at: https://github.com/citusdata/citus/pull/1656


jasonmp85:
:shipit:

PR #32, opened 2018-05-14
Use bigger instance for valgrind tests

byucesoy:
Since our tests become more memory intensive, we had errors in
valgrind tests due to available memory is not enough for valgrind.
We are starting to use bigger instance to solve this issue.

We also made some quick fixes to the scripts such as;
- adding @furkansahin to the list of people who receives test results
- removing clean section from Makefile because that was wrong in
the first place, it used to remove the actual scripts not the build
artifacts (we don't have build artifacts anyway)
- prepending install path to the path of user-data script. Adding
hardcoded path is bit hacky there but it is quick & dirty solution.
(This used to cause problems during deployment and we used
to fix this manually)

### DIFF
--- a/valgrind/Makefile
+++ b/valgrind/Makefile
@@ -7,7 +7,6 @@ VALGRIND_SCRIPTS := $(filter-out README.md Makefile,$(wildcard *))
 all:
 
 clean:
-	rm -f $(VALGRIND_SCRIPTS)
 
 install: all
 	$(INSTALL_SCRIPT) $(VALGRIND_SCRIPTS) $(DESTDIR)$(bindir)
--- a/valgrind/launch-test-instance
+++ b/valgrind/launch-test-instance
@@ -13,10 +13,10 @@ echo "Starting an instance..."
 valgrind_instance_id=$(aws ec2 run-instances \
     --image-id ami-f4cc1de2 \
     --count 1 \
-    --instance-type m3.xlarge \
+    --instance-type r3.2xlarge \
     --key-name $key_name \
     --instance-initiated-shutdown-behavior terminate \
-    --user-data file://download-test-scripts \
+    --user-data file:///usr/local/bin/download-test-scripts \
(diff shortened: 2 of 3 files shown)


valgrind/launch-test-instance
```diff
@@ -13,10 +13,10 @@ echo "Starting an instance..."
 valgrind_instance_id=$(aws ec2 run-instances \
     --image-id ami-f4cc1de2 \
     --count 1 \
-    --instance-type m3.xlarge \
+    --instance-type m5.4xlarge \
```
furkansahin: This instance type cannot be initialized without VPC set-up in the account. (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-vpc.html#vpc-only-instance-types) 
This makes the scripts not runnable for the time being. I get the following error message
```
An error occurred (VPCResourceNotSpecified) when calling the RunInstances operation: The specified instance type can only be used in a VPC. A subnet ID or network interface ID is required to carry out the request.
```

We can go with c3.8xlarge or r3.2xlarge. Both has similar RAM.

furkansahin: After changing it to r3.2xlarge, tests started successfully. Waiting for the results.

valgrind/Makefile
```diff
@@ -7,7 +7,6 @@ VALGRIND_SCRIPTS := $(filter-out README.md Makefile,$(wildcard *))
 all:
 
 clean:
```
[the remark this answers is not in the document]
byucesoy: They are required for top level Makefile.

valgrind/launch-test-instance
byucesoy: We can create a VPC and destroy it after we are done with the tests, but I guess r3 instances would also work for our purpose. So I'm inclined to just change instance type to r3.2xlarge after your tests are passed.

furkansahin:
This is all fine, after changing the instance-type to r3.2xlarge 🚢

PR #33, opened 2018-05-15
Add citus control check

lithp:
Add a script which checks that citus.control and src/backend/distriuted/citus.control are kept in sync.

Which branch should I base this off of?

### DIFF
--- a/travis/check_citus_control
+++ b/travis/check_citus_control
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+set -euo pipefail
+
+if [ ! -f citus.control ]; then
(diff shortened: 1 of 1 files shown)


jasonmp85:
Should be based on `develop` and merged to `develop`.

lithp:
@jasonmp85 I've now rebased it off `develop`. I don't think a `diff` works, because each of the files have different comments (telling you to also change the other file when you change this one). Maybe we could hard-code the expected diff output? But this script isn't very complicated.

lithp:
Closing this :)

PR #36, opened 2018-07-02
Install pg_stat_statements along with postgres

mtuncer:
Needed  this to test citus_stat_statements in travis
- loads postgres-contrib package when installing postgres from package
- compiles and installs from contrib/pg_stat_statements when installing postgres from sources

### DIFF
--- a/travis/install_custom_pg
+++ b/travis/install_custom_pg
@@ -58,7 +58,10 @@ cd postgresql
 
 make -j "${mjobs}" -s all
 make -j "${mjobs}" -s -C src/test/isolation
+make -j "${mjobs}" -s -C contrib/pg_stat_statements
+
 sudo make install
+sudo make -C contrib/pg_stat_statements install
 
 # install postgresql-common to get psql wrappers, etc.
 sudo apt-get -y -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" install postgresql-common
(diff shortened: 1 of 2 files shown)


travis/install_custom_pg
```diff
@@ -58,7 +58,10 @@ cd postgresql
 
 make -j "${mjobs}" -s all
 make -j "${mjobs}" -s -C src/test/isolation
+make -s -C contrib/pg_stat_statements
```
byucesoy: Maybe add `-j "${mjobs}"` too for consistency and slight speed-up

travis/install_pg
```diff
@@ -15,7 +15,7 @@ packages="postgresql-common libedit-dev libpam0g-dev libselinux1-dev"
 # we set PGVERSION to 10x of the Citus version when testing Citus, so
 # only install PostgreSQL proper if it's 10 or lower
 if [ "${PGVERSION//./}" -le "100" ]; then
-    packages="$packages postgresql-$PGVERSION postgresql-server-dev-$PGVERSION"
+    packages="$packages postgresql-$PGVERSION postgresql-server-dev-$PGVERSION postgresql-contrib-$PGVERSION "
```
byucesoy: Is this feature will not bbe supported in > PG10

travis/install_custom_pg
mtuncer: ok

travis/install_pg
mtuncer: moved to previous locations as you suggested

PR #37, opened 2018-07-05
add option to only check files staged for commit

thanodnl:
add diff option to `citus_indent`.

as described in the bottom I've added a `--diff` flag to only check/indent files that are staged for commit.

I had installed `citus_indent --check --quiet` as a `pre-commit` hook, which is great because you know at commit time if you made a mistake. However it started to bother me how long it took for every commit. So I updated the script to only check files to be committed. Now commits are not noticeably slower.


serprex:
Tested, will be using

PR #38, opened 2018-08-07
Use REL_11_STABLE for PG 11 builds

mtuncer:
Now stable branch is opened for PG11. master branch is now for PG 12.

We can use REL_11_STABLE for PG 11 builds.


byucesoy:
Did you create this branch from the master? If so, you should've created from develop.

mtuncer:
I was using the branch Nils created previously. 2 questions:
1 - what is the source branch (develop/master)  ?
2 - what is the target branch (develop/master) ?

byucesoy:
Both source and target should be `develop`. We use `develop` branch for new features. Then, we merge `develop` into `master` and create tag.

Issue #41, opened 2018-12-12
Enable valgrind tests to get parameters


onderkalaci:
Currently the users to to get the e-mail and the branch that is checked-out on the remote side is hard-coded. 

That requires certain workflow such as creating a branch, changing the code in the branch to use a specified branch, and finally undoing the code changes before merging the code.

It'd be good make those things parametric, externally provided to `launch-test-instance.sh` so that we don't need to follow the above hack.

onurctirtir:
We don't use this repo's valgrind tooling anymore, so closing.

PR #40, opened 2018-12-11
Add ssl and use pg 11 for Valgrind tests

onderkalaci:
Fixes #39

PR #42, opened 2019-01-09
Valgrind with custom branch

thanodnl:
This builds on #40 to allow providing custom branches and email addresses to report to passed in as arguments. This solves #41.

### DIFF
--- a/valgrind/launch-test-instance
+++ b/valgrind/launch-test-instance
@@ -16,7 +16,7 @@ valgrind_instance_id=$(aws ec2 run-instances \
     --instance-type r3.2xlarge \
     --key-name $key_name \
     --instance-initiated-shutdown-behavior terminate \
-    --user-data file:///usr/local/bin/download-test-scripts \
+    --user-data file://download-test-scripts \
     --query 'Instances[0].InstanceId' \
     --output text)
 
--- a/valgrind/launch-test-instance
+++ b/valgrind/launch-test-instance
@@ -2,6 +2,17 @@
 
 set -euo pipefail
 
+REPORT_EMAIL=${REPORT_EMAIL:-burak@citusdata.com metin@citusdata.com furkan@citusdata.com}
+POSTGRES_GITREF=${POSTGRES_GITREF:-REL_11_STABLE}
+CITUS_GITREF=${CITUS_GITREF:-master}
+
+echo "ENV:"
+echo "     REPORT_EMAIL:" $REPORT_EMAIL
+echo "  POSTGRES_GITREF:" $POSTGRES_GITREF
+echo "     CITUS_GITREF:" $CITUS_GITREF
+
+read -p "Press enter to continue"
+
 # create a key pair just for valgrind tests and store it in valgrind-test.pem
 echo "Creating key pair..."
 key_name=valgrind_$RANDOM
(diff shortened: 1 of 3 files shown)


valgrind/launch-test-instance
```diff
@@ -41,7 +53,13 @@ valgrind_instance_ip=$(aws ec2 describe-instances \
...
-ssh -o StrictHostKeyChecking=no -i $key_name.pem ubuntu@$valgrind_instance_ip "screen -d -m run-valgrind-tests"
+ssh -o StrictHostKeyChecking=no -i $key_name.pem ubuntu@$valgrind_instance_ip 'REPORT_EMAIL="${REPORT_EMAIL}" POSTGRES_GITREF="${POSTGRES_GITREF}" CITUS_GITREF="${CITUS_GITREF}" srceen -d -m run-valgrind-tests'
+
+echo login:
+echo "  " ssh -o StrictHostKeyChecking=no -i $key_name.pem ubuntu@$valgrind_instance_ip
+echo
 
 #delete the key pair after we are done with tests
-aws ec2 delete-key-pair --key-name $key_name
+echo "please run when done"
```
byucesoy: No need for this. Even if we delete the key pair automatically, the instance created with the specific key pair can be accessible with that key pair.

valgrind/run-valgrind-tests
```diff
@@ -20,31 +31,30 @@ sudo DEBIAN_FRONTEND=noninteractive apt-get install -yq \
...
-export PG_CONFIG=/usr/local/pgsql/bin/pg_config
+export PATH=$HOME/pgsql/bin:$PATH
 
 # download and install PostgreSQL
-git clone -b "REL_10_STABLE" --depth 1 git://git.postgresql.org/git/postgresql.git
+git clone -b "${POSTGRES_GITREF}" --depth 1 git://git.postgresql.org/git/postgresql.git
 cd postgresql/
-./configure --enable-cassert --enable-debug CFLAGS="-ggdb -Og -DUSE_VALGRIND"
+./configure \
+    --prefix=$HOME/pgsql \
```
byucesoy: :+1: for prefix builds.

valgrind/run-valgrind-tests
```diff
@@ -20,31 +31,30 @@ sudo DEBIAN_FRONTEND=noninteractive apt-get install -yq \
...
+git clone -b "${CITUS_GITREF}" --depth 1 https://github.com/citusdata/citus.git
 cd citus/
 ./configure
 make clean
-make -j8 -s
-sudo make install
-
-# this is necessary to start tests
-sudo chown ubuntu /usr/local/pgsql/bin/ -R
+make install -j8 -s
```
byucesoy: This is forgotten in the original script as well but we can use mjobs here too like above.

Though, I don't remember why we are adding 1 to the nproc.

valgrind/launch-test-instance
thanodnl: tru, this was for me during development, will revert back

valgrind/run-valgrind-tests
thanodnl: Good one, I will replace.

re the adding 1, this is mostly done to utilise cpu better. When you have exactly the number of jobs as cores they are often waiting on the disk as compiling is both IO and cpu intensive. By overbooking the jobs on CPU's they are more busy.

PR #44, opened 2019-07-11
Send emails to both citus and MS addresses

hanefi:
We want to receive the valgrind results in our Microsoft inboxes. We will gradually make this transition to make sure no email gets lost by being marked as spam. If we confirm that we receive the reports, only then we will drop citus mails.


hanefi:
Some extra steps need to be taken to deploy this, and thus our work here is not complete

PR #46, opened 2019-08-05
Remove citus hack as it interferes with other repos, & citus is using circle

serprex:
https://github.com/citusdata/postgresql-hll/pull/73

Is failing because after updating the sql script generation to use the C preprocessor, PG_VERSION for 9.6 was 100001

Looking over scripts, bit unsure with 9.5 being listed without 9.4 or 9.6 in [setup_apt](https://github.com/citusdata/tools/blob/develop/travis/setup_apt)

I should probably test this by submitting a pull request to postgresql-hll which uses the `fix_travis` branch

### DIFF
--- a/travis/install_pg
+++ b/travis/install_pg
@@ -10,13 +10,7 @@ if [ -n "${USE_CUSTOM_PG:-}" ]; then
 fi
 
 # always install postgresql-common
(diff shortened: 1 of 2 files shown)


serprex:
Seems using `pg_config --includedir-server` over `pg_config --includedir` has solved this

PR #50, opened 2019-09-03
Add readme for citus_dev tool

SaitTalhaNisanci:


SaitTalhaNisanci:
@thanodnl  I have updated the PR according to what you said but force pushing deleted your comment  since the hash of that line is changed.

PR #51, opened 2019-10-04
Make citus_dev compatible with any user

onlined:
If the user is not `postgres`, a new database is created with user's name on every node. This allows that `citus_dev` can be used with any user.

### DIFF
--- a/citus_dev/citus_dev
+++ b/citus_dev/citus_dev
@@ -22,6 +22,7 @@ from subprocess import call
 from subprocess import Popen, PIPE
 import os
 import sys
+import getpass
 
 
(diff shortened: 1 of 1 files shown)


JelteF:
I have put `export PGDATABASE=postgres` bashrc to solve this. I think this makes citus_dev easier to use though, so I think it's a good addition. I think this would be good to merge if the PGDATABASE environment variable is also checked, and use that instead of the username if it's set.

hanefi:
@onlined I rebased, and merged the `citus_dev` branch to `develop`, and then deleted it. Github automatically closed your PR after I deleted your base branch. Can you rebase on develop and reopen this PR?

PR #54, opened 2019-10-25
remove unused variable DESTDIR from uncrustify makefile

SaitTalhaNisanci:


hanefi:
What if there are some people in the team that wished to install the scripts in another path?

Maybe we should ask around and see if there is any logical reasoning behind having that `DESTDIR` variable

SaitTalhaNisanci:
Well, even if there is that possibility, I dont see how it is possible with this variable because with the current code, if someone sets DESTDIR to, say, `mypath` since `bindir` variable is set to `/usr/local/bin` it will be `mypath/usr/local/bin` not `mypath`.

PR #56, opened 2019-12-19
Added the new 'source' column

halilozanakgul:
Added the new 'source' column to the table. But didn't actually alter the table on the server, first need to do that

### DIFF
--- a/dashboard/pkg.jq
+++ b/dashboard/pkg.jq
@@ -82,7 +82,8 @@ def makerow(r):
                       (?<m>\\d{2})   # month
                       (?<d>\\d{2})Z$ # day";
                     "\(.y)-\(.m)-\(.d)"; "ix")),
-        .value
+        .value,
+        "PackageCloud"
     ] |
     flatten
 ;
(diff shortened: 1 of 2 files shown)


halilozanakgul:
First need to run
`ALTER TABLE download_levels ADD COLUMN source text;
UPDATE download_levels SET source='Docker Hub' WHERE os='Docker';
UPDATE download_levels SET source='RubyGems' WHERE os='ruby';

ALTER TABLE download_stats ADD COLUMN source text;
UPDATE download_stats SET source='GitHub' WHERE version='HEAD';
UPDATE download_stats SET source='Homebrew' WHERE os='macOS';
UPDATE download_stats SET source='packagecloud' WHERE source IS NULL;`

dashboard/pkg.jq
```diff
@@ -80,7 +80,8 @@ def makerow(r):
                       (?<m>\\d{2})   # month
                       (?<d>\\d{2})Z$ # day";
                     "\(.y)-\(.m)-\(.d)"; "ix")),
-        .value
+        .value,
+        "packagecloud"
```
JelteF:
```suggestion
        "PackageCloud"
```

halilozanakgul: These are how they show their names on their websites. I thought it would be more consistent to use their original names.

PR #62, opened 2020-04-01
Update packaging scripts to support all packages on a single branch

JelteF:
This is needed to support: https://github.com/citusdata/packaging/pull/441

### DIFF
--- a/automated_packaging/update_os_package.pl
+++ b/automated_packaging/update_os_package.pl
@@ -69,12 +69,12 @@ $curTime = time();
 `sed -i 's/^pkglatest.*/pkglatest=$VERSION.citus-1/g' pkgvars`;
 
 # Based on the repo, update the package related variables
-if ( $DISTRO_VERSION eq "redhat" || $DISTRO_VERSION eq "microsoft") {
+if ( $DISTRO_VERSION eq "redhat" || $DISTRO_VERSION eq "microsoft" || $DISTRO_VERSION eq "all") {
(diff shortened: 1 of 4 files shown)


automated_packaging/update_os_package.pl
```diff
@@ -69,12 +69,12 @@ sub get_changelog_for_debian {
 `sed -i 's/^pkglatest.*/pkglatest=$VERSION.citus-1/g' pkgvars`;
 
 # Based on the repo, update the package related variables
-if ( $DISTRO_VERSION eq "redhat" || $DISTRO_VERSION eq "microsoft") {
+if ( $DISTRO_VERSION eq "redhat" || $DISTRO_VERSION eq "microsoft" || $DISTRO_VERSION eq "all") {
```
hanefi, line 72: I think we should document the allowed `$DISTRO_VERSION` values after we merge this.

We may either introduce some documentation like we have in citus_package https://github.com/citusdata/tools/blob/4cf2cd1c5eeda887d4ed3e3cfb66ace69e15f468/packaging/citus_package#L405-L560

This idea may be outside the scope of this PR, just sharing my opinion

onurctirtir: Or before introducing that, we should at least introduce this "all" parameter to our word doc

JelteF: yeah, once we merged all the PRs I think we should update our worddoc to use the "all" parameter.

PR #69, opened 2020-05-06
Support pgautofailover in update_os_package.pl

JelteF:
This makes releasing pgautofailover much less involved. This PR has been created using these changes: https://github.com/citusdata/packaging/pull/469


automated_packaging/update_os_package.pl
```diff
@@ -9,10 +9,24 @@
 # Name of the repo is represented differently on logs and repos
 my $github_repo_name = "citus";
 my $log_repo_name = "Citus";
+my $version_suffix = ".citus";
```
hanefi, line 12: We ended up dropping this suffix for pg_auto_failover. Is there any change we need to make for that

hanefi: We overwrite the value below. No need to fix anything

PR #76, opened 2020-05-11
Set all automated PRs as draft

hanefi:
When we use the scripts to update branches, and open pull requests there can be issues with the resulting PRs occasionally.

I believe the owner of a packaging task should review all automated PRs and set the PR as ready only after confirming that the contents are fine


hanefi:
`s/true/True` should fix the parsing issues in the curl post parameters.

onurctirtir:
iirc, https://github.com/citusdata/tools/pull/76#issuecomment-647385193 didn't work

What is the status of this pr @hanefi ?

PR #78, opened 2020-06-14
Add support for Postgres 13

lfittl:
This is modeled after d9eb817d1d66d22de561d6d78f389065517b3bde, with
a few less changes because the logic was simplified back then.

Needed for https://github.com/citusdata/pg_auto_failover/pull/297 (which adds Postgres 13 support to pg_auto_failover)

### DIFF
--- a/travis/setup_apt
+++ b/travis/setup_apt
@@ -12,11 +12,11 @@ sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 7FCC7D46ACCC4CF8
 sudo rm -f /etc/apt/sources.list.d/google-chrome*
 sudo rm -f /etc/apt/sources.list.d/pgdg.list
 
-# add the PostgreSQL 12 repository
-sudo sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -cs)-pgdg main 12" >> /etc/apt/sources.list.d/postgresql.list'
+# add the PostgreSQL 13 repository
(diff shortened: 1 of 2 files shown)


travis/setup_apt
```diff
@@ -12,11 +12,11 @@ sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 7FCC7D46ACCC4CF8
 sudo rm -f /etc/apt/sources.list.d/google-chrome*
 sudo rm -f /etc/apt/sources.list.d/pgdg.list
 
-# add the PostgreSQL 12 repository
-sudo sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -cs)-pgdg main 12" >> /etc/apt/sources.list.d/postgresql.list'
+# add the PostgreSQL 13 repository
+sudo sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -cs)-pgdg main 13" >> /etc/apt/sources.list.d/postgresql.list'
```
hanefi, line 16: We can drop `13` as PGDG no longer has seperate repositories for different PG versions.
```suggestion
sudo sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -cs)-pgdg main" >> /etc/apt/sources.list.d/postgresql.list'
```

See https://www.postgresql.org/download/linux/debian/ for latest repo installation instructions

lfittl-msft: I think thats not correct for versions that aren't fully released yet - see https://wiki.postgresql.org/wiki/Apt/FAQ#I_want_to_try_the_beta_version_of_the_next_PostgreSQL_release

PR #82, opened 2020-11-09
Skip pg13 for el/ol 6

onurctirtir:
Proper way to disable pg13 package builds for rpm based distros is to modify `pkgvars` files from `all-x` branches.
However, due to the infrastructure that our tooling uses, it would cost much more.
For simplicity, disable them here.


hanefi (approved):
This is safe to merge. However, we need to keep in mind to revert the exclusion rules here once the packages for el6 and ol6 are ready.

:shipit:

PR #83, opened 2020-11-13
Skip pg13 for some distros in build_x_release scripts too

onurctirtir:
Complete the changes done in https://github.com/citusdata/tools/pull/82.

Previously, as `latestpg` that our scripts think that we should have in PackageCloud becomes 13 so script was trying to push nightly images.
However, as they already exists in PackageCloud, ci was reporting error.

### DIFF
--- a/travis/build_new_nightly
+++ b/travis/build_new_nightly
@@ -15,7 +15,7 @@ if [ "${TARGET_PLATFORM:-}" = "" ]; then
 fi
 
 PLATFORM_TYPE="$(echo "$TARGET_PLATFORM" | cut -d "/" -f1)"
-PLATFORM_VERSION="$(echo "$TARGET_PLATFORM" | cut -d "/" -f2 | cut -d "/" -f1)"
+PLATFORM_VERSION="$(echo "$TARGET_PLATFORM" | cut -d "/" -f2)"
 pkgauth="${PACKAGECLOUD_API_TOKEN}:"
 hubauth="Authorization: token ${GITHUB_TOKEN}"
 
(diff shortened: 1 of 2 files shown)


travis/build_new_nightly
```diff
@@ -14,7 +14,8 @@ if [ "${TARGET_PLATFORM:-}" = "" ]; then
     exit $badusage
 fi
 
-PLATFORM_TYPE="${TARGET_PLATFORM%%/*}"
+PLATFORM_TYPE="$(echo "$TARGET_PLATFORM" | cut -d "/" -f1)"
+PLATFORM_VERSION="$(echo "$TARGET_PLATFORM" | cut -d "/" -f2 | cut -d "/" -f1)"
```
hanefi, line 18: My bash scripting experience is rather limited, so please bear with me.

1. Why do you not split the string using `/` as a delimiter?
2. IIRC `TARGET_PLATFORM` variables contain a single `/` character. The second cut operation would not actually do anything, right?

onurctirtir: 1 - doing so actually
2 - true, I will remove the second one

hanefi:
```bash
IFS=/; read -a TARGET <<<"$TARGET_PLATFORM"
PLATFORM_TYPE = $TARGET[0]
PLATFORM_VERSION = $TARGET[1]
```

onurctirtir: https://github.com/citusdata/tools/pull/83/commits/6b687fbc521536a9c4e51558a1f21933627b4f44

hanefi (approved):
:shipit: 

We'll see the status of next cron builds on CI. I hope these changes will fix everything

PR #109, opened 2021-06-12
Add common utils for further developments

gurkanindibay:


test1e58b070-b0bb-4f78-9ae4-3672bd131097
```diff
@@ -0,0 +1 @@
+Test content
```
onurctirtir, line 1: let's remove those three test[.*] files if we don't need them

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +321,114 @@ def remove_cloned_code(exec_path: str):
...
+    except subprocess.TimeoutExpired:
+        raise ValueError(
+            f"Error while getting key. Most probably packaging key is stored with password. "
+            f"Please remove the password when storing key with fingerprint {fingerprint}")
+
+
+def get_public_gpg_key(fingerprint: str) -> str:
+    gpg = gnupg.GPG()
+    ascii_armored_public_keys = gpg.export_keys(fingerprint)
+    return base64.b64encode(ascii_armored_public_keys.encode("ascii")).decode("ascii")
```
onurctirtir, line 391: I'm wondering why we are encoding and decoding those keys ?

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +321,114 @@ def remove_cloned_code(exec_path: str):
...
+    except subprocess.TimeoutExpired:
+        raise ValueError(
+            f"Error while getting key. Most probably packaging key is stored with password. "
+            f"Please remove the password when storing key with fingerprint {fingerprint}")
+
+
+def get_secret_key_by_fingerprint_with_password(fingerprint: str, passphrase: str) -> str:
+    try:
+        gpg = gnupg.GPG()
+        private_key = gpg.export_keys(fingerprint, True, passphrase=passphrase)
```
onurctirtir, line 379: 1) & 2) I think comments for `get_secret_key_by_fingerprint` apply to this function too.

Also, minor: Let's specify what are we setting to `True` (like X=True) since readers might not know about gpg's API

packaging_automation/tests/test_utils.py
```diff
@@ -1,4 +1,8 @@
 import difflib
+from ..common_tool_methods import *
```
onurctirtir, line 2: I think it would be nicer to explicitly specify what we are importing

packaging_automation/tests/test_common_tool_methods.py
```diff
@@ -1,201 +1,192 @@
...
+        commit_message = f"Test message for {test_branch_name}"
+        run(f"git commit -m '{commit_message}'")
+        assert get_last_commit_message(os.getcwd()) == f"{commit_message}\n"
+    finally:
+        run(f"git checkout {current_branch_name}")
+        run(f"git branch -D {test_branch_name}")
+
+
+
+
```
onurctirtir, line 90: minor: I think we can remove a few of those new lines, by following what we are normally doing between other functions

packaging_automation/common_tool_methods.py
```diff
@@ -32,10 +35,33 @@ def release_all_repos():
...
+    project_version_details = get_version_details(project_version)
+    return f'{project_version_details["major"]}.{project_version_details["minor"]}'
+
+
+def get_version_number(version: str, fancy: bool, fancy_release_count: int) -> str:
+    fancy_suffix = f"-{fancy_release_count}" if fancy else ""
+    return f"{version}{fancy_suffix}"
+
+
+def get_version_number_with_project_name(project_name: str, version: str, fancy: bool, fancy_release_count: int) -> str:
```
onurctirtir, line 60: maybe `fancy_release_count` -> `fancy_release_number`

packaging_automation/common_tool_methods.py
```diff
@@ -32,10 +35,33 @@ def release_all_repos():
...
 def get_spec_file_name(project_name: str) -> str:
     return f"{project_name}.spec"
 
 
+def get_minor_project_version(project_version: str) -> str:
+    project_version_details = get_version_details(project_version)
+    return f'{project_version_details["major"]}.{project_version_details["minor"]}'
+
+
+def get_version_number(version: str, fancy: bool, fancy_release_count: int) -> str:
```
onurctirtir, line 55: Will we ever call either of `get_version_number` & `get_version_number_with_project_name` with `fancy = False`.
If not, I think we can remove `fancy: bool` params and rename those two functions as such:

`get_version_number` -> `get_fancy_version`
`get_version_number_with_project_name` -> `get_fancy_version_with_project_name`

packaging_automation/common_tool_methods.py
```diff
@@ -138,7 +169,7 @@ def get_prs_for_patch_release(repo: Repository.Repository, earliest_date: dateti
     # filter pull requests according to given time interval
     filtered_pull_requests = list()
     for pull_request in pull_requests:
-        if not pull_request.is_merged():
+        if not pull_request.merged_at:
```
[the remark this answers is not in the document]
gurkanindibay: Because everytime is_merged is called, an api request is done which will cause API limit exceed

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +321,114 @@ def remove_cloned_code(exec_path: str):
         except:
             print(f"Some files could not be deleted in directory {exec_path}. "
                   f"Please delete them manually or they will be deleted before next execution")
+
+
+def process_docker_template_file(project_version: str, templates_path: str, template_file_path: str):
+    minor_version = get_minor_project_version(project_version)
+    env = get_template_environment(templates_path)
+    template = env.get_template(template_file_path)
+    return f"{template.render(project_version=project_version, project_minor_version=minor_version)}\n"
```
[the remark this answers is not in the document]
gurkanindibay: Fixed Added a comment and unit test

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +321,114 @@ def remove_cloned_code(exec_path: str):
...
+
+
+def delete_gpg_key_by_name(name: str):
+    try:
+        while True:
+            key_id = get_gpg_fingerprint_from_name(name)
+            run(f"gpg --batch --yes --delete-secret-key {key_id}")
+            run(f"gpg --batch --yes --delete-key {key_id}")
+    except ValueError:
+        print(f"Key for the name {name} does not exist")
```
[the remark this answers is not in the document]
gurkanindibay: I split into two methods and write comment inside the methods. 
There could be more than one key with the same name so delete command is executed until no key remains i.e. exception is thrown

packaging_automation/common_tool_methods.py
gurkanindibay: deleted since it is not used

packaging_automation/common_tool_methods.py
```diff
@@ -78,6 +104,11 @@ def run(command, *args, **kwargs):
     return result
 
 
+def run_with_output(command, *args, **kwargs):
+    result = subprocess.run(command, *args, check=True, shell=True, stdout=subprocess.PIPE, **kwargs)
```
SaitTalhaNisanci, line 108: As far as I know, using `shell=True` should be avoided when possible because this would mean you could execute a lot of unsafe things. Maybe see https://stackoverflow.com/questions/3172470/actual-meaning-of-shell-true-in-subprocess

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +321,114 @@ def remove_cloned_code(exec_path: str):
...
+    files = list()
+    for (dirpath, dirnames, filenames) in os.walk(rpm_dir_path):
+        files += [os.path.join(dirpath, file) for file in filenames]
+    rpm_files = filter(lambda file_name: file_name.endswith("rpm"), files)
+    for file in rpm_files:
+        if not is_rpm_file_signed(f"{file}"):
+            raise ValueError(f"File {file} is not signed or there is a signature check problem")
+
+def remove_prefix(text, prefix):
+    return text[text.startswith(prefix) and len(prefix):]
```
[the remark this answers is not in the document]
SaitTalhaNisanci: I think what is suggested is more understandable because it is simpler. For example what does it mean when `startswith` returns `False`? (I guess if we were to go with the suggestion, we wouldn't need to answer this)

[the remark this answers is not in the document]
gurkanindibay: Fixed with a small change

packaging_automation/common_tool_methods.py
gurkanindibay: Fixed

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +323,171 @@ def remove_cloned_code(exec_path: str):
...
+    # When getting gpg key if gpg key is stored with password and if given passphrase is wrong, timeout exception is
+    # thrown.
+    gpg = gnupg.GPG()
+    try:
+        private_key = gpg.export_keys(fingerprint, secret=True, passphrase=passphrase)
+    except subprocess.TimeoutExpired:
+        raise ValueError(
+            f"Error while getting key. Most probably packaging key is stored with password. "
+            f"Please check the password and try again")
+    return base64.b64encode(private_key.encode("ascii")).decode("ascii")
```
onurctirtir, line 442: For `get_secret_key_by_fingerprint` & `get_secret_key_by_fingerprint_with_password`

I'm still not much convinced about why it's possible to get `TimeoutExpired` exceptions, instead of plain errors.

Can you look into that a bit more to see if there is a way to get secret keys without dealing with `TimeoutExpired` exceptions ?
It would also be nice if you can share your findings here so that I can also help solving this issue.

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +323,171 @@ def remove_cloned_code(exec_path: str):
...
+    run(f"gpg --export -a {fingerprint} >rpm_public.key")
+    run("rpm --import rpm_public.key")
+    os.remove("rpm_public.key")
+
+
+def delete_rpm_key_by_name(key_name: str):
+    result = run_with_output("rpm -q gpg-pubkey --qf %{NAME}-%{VERSION}-%{RELEASE}__%{SUMMARY}-!!")
+    output = result.stdout.decode("ascii")
+    if result.stderr:
+        print(f"Error!!!{output}")
```
onurctirtir, line 455: Here we check if we have something in `stderr` but printing the `stdout`.
Maybe we should do something like that:

```suggestion
    if result.stderr:
        err = result.stderr.decode("ascii")  
        print(f"Error: {err}")
```

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +323,171 @@ def remove_cloned_code(exec_path: str):
...
+    else:
+        # Get all the lines which includes rpm keys and get the first item which is rpm key to be used to delete
+        # the key .Example line gpg-pubkey-fd431d51-4ae0493b__gpg(Red Hat, Inc. (release key 2) <security@redhat.com>)
```
onurctirtir, line 458: Can we strictly rely on the fact that line will always have `__` /`!!` as the separator between key and the rest ?
Can you share a link that explicitly declares this format ?

packaging_automation/common_tool_methods.py
onurctirtir: Personally, I don't quite like making a function no-op depending on the parameter.
So I think we can do something like that for pg-autofailover, what do you think ?

```python
if pg_autofailover:
    version = get_fancy_version(version)
    # or version = get_fancy_version_with_project_name(version )
```

If that makes sense to you too, then I think we can remove `fancy` param and rename those two functions.

packaging_automation/common_tool_methods.py
```diff
@@ -78,6 +105,12 @@ def run(command, *args, **kwargs):
     return result
 
 
+def run_with_output(command, *args, **kwargs):
+    result = subprocess.run(command, *args, shell=True, capture_output=True,
+                            **kwargs)
```
onurctirtir, line 110: minor:

```suggestion
    result = subprocess.run(command, *args, shell=True, capture_output=True, **kwargs)
```

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +323,171 @@ def remove_cloned_code(exec_path: str):
         except:
             print(f"Some files could not be deleted in directory {exec_path}. "
                   f"Please delete them manually or they will be deleted before next execution")
+
+
+def process_template_file(project_version: str, templates_path: str, template_file_path: str):
+    '''Process template files and return the processed template '''
```
onurctirtir, line 329: Can we put some comments to clarify a few things here ? Like:

* What is a template file and for which purposes we use them ?
* What does "process" mean here ? What do we do by calling "render" function ?

packaging_automation/common_tool_methods.py
onurctirtir: Let's do that for `get_version_number` too

packaging_automation/common_tool_methods.py
onurctirtir: Looks like we still set `shell` to `True` by default, so this doesn't seem to be resolved yet.

packaging_automation/common_tool_methods.py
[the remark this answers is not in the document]
gurkanindibay: Fixed

packaging_automation/common_tool_methods.py
gurkanindibay: fancy version number is needed to get fancy version. I got as parameter

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +323,171 @@ def remove_cloned_code(exec_path: str):
...
+
+def write_to_file(content: str, dest_file_name: str):
+    with open(dest_file_name, "w") as writer:
+        writer.write(content)
+
+
+def get_gpg_fingerprint_from_name(name: str):
+    '''Gets gpg key fingerprint from the key name. GPG fingerprint is unique parameter. However,
+    It is hard to track and reference in the code. When signing packages, gpg keys are referenced by name
+    and this function is used to get the unique identifier so fingerprint from name'''
```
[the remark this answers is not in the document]
gurkanindibay: Fixed as suggested

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +321,114 @@ def remove_cloned_code(exec_path: str):
...
+    line_found = False
+    for line in lines:
+        if line.startswith("uid") and name in line:
+            line_found = True
+            break
+        counter = counter + 1
+    if not line_found:
+        raise ValueError(f"Key with the name {name} could not be found ")
+    else:
+        return lines[counter - 1].strip()
```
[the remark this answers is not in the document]
gurkanindibay: I changed implementation to get the list of all keys with the same name and simplified the implementation and got rid of both counter and has_line

packaging_automation/common_tool_methods.py
gurkanindibay: Sorry but my IDE deleted the code I wrote before I commit. :( I'm trying to solve that problem. Now I handled this again. Please check it

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +323,171 @@ def remove_cloned_code(exec_path: str):
...
+    line_found = False
+    for line in lines:
+        if line.startswith("uid") and name in line:
+            line_found = True
+            break
+        counter = counter + 1
+    if not line_found:
+        return ""
+    else:
+        return lines[counter - 1].strip()
```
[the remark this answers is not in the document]
gurkanindibay: Simplified it

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +323,171 @@ def remove_cloned_code(exec_path: str):
...
+            print(f"Public key with the name {name} deleted")
+            counter = counter + 1
+        elif output.returncode == 2:  # key not exist
+            print(f"Key for the name {name} does not exist")
+            break
+        else:
+            print(f"Error {output.stderr.decode('ascii')}")
+            break
+        if counter > 0:
+            print(f"{counter} public  keys deleted")
```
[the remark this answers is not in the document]
gurkanindibay: unified the implementation. output.returncode == 2 is important because key_id is the same for both private and public keys but sometimes one of them may be missing. To make sure to delete all the keys we need to continue on key non-existence case

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +323,171 @@ def remove_cloned_code(exec_path: str):
...
+            print(f"Private key with the name {name} deleted")
+            counter = counter + 1
+        elif output.returncode == 2:  # key not exist
+            print(f"Key for the name {name} does not exist")
+            break
+        else:
+            print(f"Error {output.stderr.decode('ascii')}")
+            break
+        if counter > 0:
+            print(f"{counter} private  keys deleted")
```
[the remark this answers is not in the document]
gurkanindibay: Fixed

packaging_automation/common_tool_methods.py
gurkanindibay: Parameterized them as key_separator and line_separator. We are using this statements in the command

packaging_automation/common_tool_methods.py
gurkanindibay: Fixed

packaging_automation/common_tool_methods.py
gurkanindibay: Added some more comments. I hope this one is enough

packaging_automation/common_tool_methods.py
gurkanindibay: Fixed

packaging_automation/common_tool_methods.py
gurkanindibay: These parameters are defined by me. It is not a general thing. So there could not be a document about it. __ and !! character series are just seperators

gurkanindibay: image

gurkanindibay: I can handle with some escape characters one is required for bash another one is required for python

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +335,151 @@ def remove_cloned_code(exec_path: str):
...
+    # gpg.export_keys needs to get password. Otherwise it gives error. Dummy password is only to bypass parameter
+    # error
+    dummy_password = "123"
+
+    return get_secret_key_by_fingerprint_with_password(fingerprint, dummy_password)
+
+
+def get_secret_key_by_fingerprint_with_password(fingerprint: str, passphrase: str) -> str:
+    # When getting gpg key if gpg key is stored with password and if given passphrase is wrong, timeout exception is
+    # thrown.
```
onurctirtir, line 417: outdated comment, we don't expect timeout now.

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +335,151 @@ def remove_cloned_code(exec_path: str):
...
+
+
+def delete_rpm_key_by_name(key_name: str):
+    line_separator = "!!"
+    key_separator = "__"
+    result = run_with_output(
+        "rpm -q gpg-pubkey --qf %{NAME}-%{VERSION}-%{RELEASE}" + key_separator + "%{SUMMARY}-" + line_separator)
+    output = result.stdout.decode("ascii")
+    if result.stderr:
+        print(f"Error!!!{result.stderr.decode('ascii')}")
```
onurctirtir, line 450:
```suggestion
        print(f"Error: {result.stderr.decode('ascii')}")
```

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +335,151 @@ def remove_cloned_code(exec_path: str):
         except:
             print(f"Some files could not be deleted in directory {exec_path}. "
                   f"Please delete them manually or they will be deleted before next execution")
+
+
+def process_template_file(project_version: str, templates_path: str, template_file_path: str):
+    ''' This function gets the template files, changes tha parameters inside the file and return the output.
```
onurctirtir, line 341:
```suggestion
    ''' This function gets the template files, changes the parameters inside the file and returns the output.
```

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +335,151 @@ def remove_cloned_code(exec_path: str):
...
+        Template files are stored under packaging_automation/templates and these files includes parametric items in the
+        format of {{parameter_name}} This function is used creating while docker files, pgxn files which includes
+        project_name as parameter. Example usage is in test_common_tool_methods/test_process_template_file.
+        Jinja2 is used as template engine and render function gets the file change parameters in the file with the given
+        parameters as input and returns the output '''
```
onurctirtir, line 341:
```suggestion
    ''' This function gets the template files, changes the parameters inside the file and returns the output.
        Template files are stored under packaging_automation/templates and these files include parametric items in the
        format of {{parameter_name}}. This function is used while creating docker files and pgxn files which include
        "project_name" as parameter. Example usage is in "test_common_tool_methods/test_process_template_file".
        Jinja2 is used as the template engine and render function gets the file change parameters in the file with the given
        input parameters and returns the output'''
```

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +330,151 @@ def remove_cloned_code(exec_path: str):
...
+    delete_private_gpg_key_by_name(name)
+    delete_public_gpg_key_by_name(name)
+
+
+def get_secret_key_by_fingerprint_without_password(fingerprint: str) -> str:
+    # gpg.export_keys needs to get password. Otherwise it gives error. Dummy password is only to bypass parameter
+    # error
+    dummy_password = "123"
+
+    return get_secret_key_by_fingerprint_with_password(fingerprint, dummy_password)
```
onurctirtir, line 407: could you check gpg manuel to see if there is a better way to do that, instead of passing a dummy password ?

packaging_automation/common_tool_methods.py
gurkanindibay: Fixed with the soluyion you shared on Teams

packaging_automation/common_tool_methods.py
gurkanindibay: Fixed

packaging_automation/common_tool_methods.py
gurkanindibay: Fixed

packaging_automation/common_tool_methods.py
gurkanindibay: Fixed by upgrading the library

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +323,154 @@ def remove_cloned_code(exec_path: str):
...
+
+def write_to_file(content: str, dest_file_name: str):
+    with open(dest_file_name, "w") as writer:
+        writer.write(content)
+
+
+def get_gpg_fingerprints_by_name(name: str) -> List[str]:
+    '''Returns GPG fingerprint by its unique key name. We use this function to determine the fingerprint that
+       we should use when signing packages'''
+    result = subprocess.run(shlex.split(f"gpg --list-keys "), check=True, stdout=subprocess.PIPE)
```
onurctirtir, line 349: minor:

```suggestion
    result = subprocess.run(shlex.split(f"gpg --list-keys"), check=True, stdout=subprocess.PIPE)
```

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +323,154 @@ def remove_cloned_code(exec_path: str):
...
+    return finger_prints
+
+
+def delete_gpg_key_by_name(name: str, key_type: GpgKeyType):
+    keys = get_gpg_fingerprints_by_name(name)
+
+    # There could be more than one key with the same name. While statement is used to delete all the public keys
+    # until no key remains (i.e. key_id is empty).
+    # Public and private keys are stored with the same fingerprint. In some cases one of them may not be exist.
+    # Therefore non-existence case is possible
```
onurctirtir, line 367: outdated comment
```suggestion
```

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +323,154 @@ def remove_cloned_code(exec_path: str):
...
+    for key_id in keys:
+        delete_command = f"gpg --batch --yes --delete-key {key_id}" if key_type == GpgKeyType.public \
+            else f"gpg --batch --yes --delete-secret-key {key_id}"
```
onurctirtir, line 370: for readability, maybe not inline but have something like this:

```python
if key_type == GpgKeyType.public:
  delete_command = 
else if:
  delete_command = 
else:
  # maybe throw an error to tell that we got an unexpected key type
```

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +323,154 @@ def remove_cloned_code(exec_path: str):
...
+        output = run_with_output(delete_command)
+        if output.returncode == 0:
+            print(f"{key_type.name.capitalize()} key with the id {key_id} deleted")
+        #
```
onurctirtir, line 374:
```suggestion
```

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +323,154 @@ def remove_cloned_code(exec_path: str):
...
+        elif output.returncode == 2:  # Key does not exist in keyring
```
onurctirtir, line 375: minor style comment, I'd like to see comments within the conditional blocks, but it's up to you:

```suggestion
        # Key does not exist in keyring
        elif output.returncode == 2:
```

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +323,154 @@ def remove_cloned_code(exec_path: str):
...
+    run("rpm --import rpm_public.key")
+    os.remove("rpm_public.key")
+
+
+def delete_rpm_key_by_name(summary: str):
+    rpm_keys = get_rpm_keys()
+    for key in rpm_keys:
+        if has_rpm_key_given_summary(key, summary):
+            run(f"rpm -e {key}")
+            print(f"RPM key with id {key} was deleted")
```
onurctirtir, line 437: Why do we delete by summary ? Isn't it safer to delete by key name ?

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +323,154 @@ def remove_cloned_code(exec_path: str):
...
+def get_rpm_keys():
+    result = run_with_output("rpm -q gpg-pubkey")
+    if result.stderr:
+        raise ValueError(f"Error:{result.stderr.decode('ascii')}")
+    output = result.stdout.decode("ascii")
+    key_lines = output.splitlines()
+    return key_lines
+
+
+def has_rpm_key_given_summary(key: str, summary: str):
```
onurctirtir, line 449: Also, if we will keep this function (see the comment above), maybe `has_rpm_key_given_summary` -> `rpm_key_matches_summary`.

packaging_automation/common_tool_methods.py
```diff
@@ -290,3 +323,154 @@ def remove_cloned_code(exec_path: str):
...
+    result = run_with_output("rpm -q " + key + " --qf  '%{SUMMARY}' ")
```
onurctirtir, line 450: very minor:

```suggestion
    result = run_with_output("rpm -q " + key + " --qf  '%{SUMMARY}'")
```

packaging_automation/common_tool_methods.py
```diff
@@ -32,10 +36,39 @@ def release_all_repos():
...
+
+def get_version_number(version: str, fancy: bool, fancy_release_count: int) -> str:
+    fancy_suffix = f"-{fancy_release_count}" if fancy else ""
+    return f"{version}{fancy_suffix}"
+
+
+def get_version_number_with_project_name(project_name: str, version: str, fancy: bool,
+                                         fancy_release_number: int) -> str:
+    fancy_suffix = f"-{fancy_release_number}" if fancy else ""
+    return f"{version}.{project_name}{fancy_suffix}"
```
onurctirtir, line 69: 1) `_number` word doesn't sound much suitable for those func names, since we are indeed returning strings in both.

Can you come up with better names for those functions ?

Maybe `get_version_number` -> `process_version_string` & `get_version_number_with_project_name` -> `process_version_with_project_name`

2) `fancy_release_count` -> `fancy_release_number`

onurctirtir (approved):
LGTM. Conditionally approving, mostly left some minor comments.
Please make sure to address them.

Also, I couldn't quite understand why are we deleting the keys by "SUMMARY".
To me, it doesn't sound like a safe approach.
https://github.com/citusdata/tools/pull/109#discussion_r655205925

packaging_automation/common_tool_methods.py
gurkanindibay: Instead of while I used for in the comment. Other than this issue, I think this comment is still applicable

packaging_automation/common_tool_methods.py
gurkanindibay: Fixed

packaging_automation/common_tool_methods.py
gurkanindibay: Fixed

packaging_automation/common_tool_methods.py
gurkanindibay: Fixed

packaging_automation/common_tool_methods.py
gurkanindibay: Fixed

packaging_automation/common_tool_methods.py
onurctirtir: I meant we don't expect timeout exception now, so we can just delete this comment.

packaging_automation/common_tool_methods.py
onurctirtir: I think we just delete by key ?

gurkanindibay: Could you check it again. Because I didn't approve the review. It's my fault

packaging_automation/common_tool_methods.py
gurkanindibay: I already deleted the comment now you shouldn't see the comment

onurctirtir: still seeing the comment here:
https://github.com/citusdata/tools/pull/109/files#diff-17da5866902b4d28bf41a5e67c26fb52a65506574f43ab6c09e42db2102d8292R24-R25

PR #115, opened 2021-06-21
Fix --init-with and add --init-worker-with

JelteF:
This fixes easy initialization of a cluster, and it makes it easy to initialize
a worker with some different commands.


thanodnl (approved):
LGTM

PR #120, opened 2021-06-28
Update problematic python lib versions detected by dependabot

gurkanindibay:
cd packaging_automation
pip-compile

### DIFF
--- a/packaging_automation/requirements.in
+++ b/packaging_automation/requirements.in
@@ -1,17 +1,17 @@
 wheel
 docker
 GitPython
-Jinja2
+Jinja2 ==2.11.3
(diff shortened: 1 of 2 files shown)


hanefi (approved):
👍🏽

hanefi (changes requested):
I think we should not start hardcoding package versions in `requiremenst.in` file, instead keep the versions only on the `requirements.txt` if possible.

gurkanindibay:
I try to find a soltion not to hardcode but Maybe give a range in requirements.in. however changing requirements.txt is not a good option since requreiements.txt is a "compiled" file

PR #125, opened 2021-07-13
Update package properties general fixes

gurkanindibay:
This PR includes multiple fixes for issues I faced while I was using update_package_properties.py script for ther release of citus 9.4.5, citus 9.5.6 and pgaf 1.6.1. 
Issues fixed in PR
* Script didn't support citus-enterprise, pgaf and pgaf-enterprise. I added support by adding templates and created a modular structure to add more project in the future
* This script had a different structure other than other scripts like prepare_release, update_docker. A bash script was introduced to excute the script but it made it difficult to execute. I removed the bash script and now it is being executed with python -m command like others
* Prepare_release script is all-in-one script which clones the project to be configured for the release. This script now have the similar feature which makes it easy to use and less error prone.


packaging_automation/upload_to_package_cloud.py
```diff
@@ -116,13 +116,14 @@ def package_exists(package_cloud_token: str, repo_owner: str, repo_name: str, pa
...
     parser.add_argument('--output_file_path', required=True)
+    parser.add_argument('--exec_folder_path', required=True)
 
     args = parser.parse_args()
 
     multiple_return_value = upload_files_in_directory_to_package_cloud(args.output_file_path,
                                                                        args.platform,
                                                                        args.package_cloud_api_token,
-                                                                       args.repository_name)
+                                                                       args.repository_name, args.exec_folder_path)
```
onurctirtir, line 126: what is the usage of this parameter ?

packaging_automation/tests/test_update_package_properties.py
```diff
@@ -188,11 +190,12 @@ def test_update_all_changes():
     copyfile(spec_file, spec_file_copy)
 
     try:
-        package_properties_param = PackagePropertiesParams(project_name=PROJECT_NAME,
-                                                           project_version=PROJECT_VERSION, fancy=True,
-                                                           fancy_version_number=1,
-                                                           name_surname=NAME_SURNAME, microsoft_email=MICROSOFT_EMAIL,
-                                                           changelog_date=CHANGELOG_DATE)
+        package_properties_param = PackagePropertiesParams(
+            supported_project=DEFAULT_CHANGELOG_PARAM_FOR_TEST.supported_project,
```
onurctirtir, line 194: what does `supported_project` mean here ?

packaging_automation/templates/citus-enterprise.spec.tmpl
```diff
@@ -0,0 +1,347 @@
...
+Summary:	PostgreSQL-based distributed RDBMS
+Name:		%{sname}%{?pkginfix}_%{pgmajorversion}
+Provides:	citus_%{pgmajorversion}
+Conflicts:	citus_%{pgmajorversion}
+Version:	{{rpm_version}}
+Release:	1%{dist}
+License:	Commercial
+Group:		Applications/Databases
+Source0:	https://github.com/citusdata/{{project_name}}/archive/v{{version}}.tar.gz
+URL:		https://github.com/citusdata/{{project_name}}
```
onurctirtir, line 18: I think we should not have `{{project_name}}` vars in template files, since we can anyway hardcode it.

Can you go over `.tmpl` files and replace occurrences of `{{project_name}}` with the actual project names ?

packaging_automation/update_package_properties.py
```diff
@@ -2,36 +2,62 @@
...
+                                                 changelog_project_name="pg_auto_failover-enterprise",
+                                                 packaging_branch="all-pgautofailover-enterprise")
 
 
 @parameter_validation
 def is_project_changelog_header(header: str):
     if not header:
         raise ValueError("header should be non-empty and should not be None")
-    if not re.match(r"^### \w+\sv\d+\.\d+\.\d+\s\(\w+\s\d+,\s\d+\)\s###$", header):
+    if not re.match(r"^### \w+[-]?\w+\sv\d+\.\d+\.\d+\s\(\w+\s\d+,\s\d+\)\s###$", header):
```
onurctirtir, line 53: can we have comment to make it clear what does that regex catches, like:
`# an example matching string is "citus-X Y foo-bar"`

packaging_automation/update_package_properties.py
```diff
@@ -2,36 +2,62 @@
...
+@dataclass()
+class ProjectDetails:
+    name: str
+    version_suffix: str
+    github_repo_name: str
+    changelog_project_name: str
+    packaging_branch: str
+
+
+class SupportedProjects(Enum):
```
onurctirtir, line 34: minor: I think we should remove "s" in the end:

```suggestion
class SupportedProject(Enum):
```

packaging_automation/update_package_properties.py
```diff
@@ -2,36 +2,62 @@
...
-    if not re.match(r"^### \w+\sv\d+\.\d+\.\d+\s\(\w+\s\d+,\s\d+\)\s###$", header):
+    if not re.match(r"^### \w+[-]?\w+\sv\d+\.\d+\.\d+\s\(\w+\s\d+,\s\d+\)\s###$", header):
         raise ValueError(
             f"changelog header is in invalid format. Actual:{header} Expected: ### citus v8.3.3 (March 23, 2021) ### ")
 
 
 @dataclass
 class PackagePropertiesParams:
-    project_name: str
+    supported_project: SupportedProjects
```
onurctirtir, line 60:
```suggestion
    project: SupportedProjects
```

packaging_automation/update_package_properties.py
```diff
@@ -40,22 +66,36 @@ class PackagePropertiesParams:
     latest_changelog: str = ""
     changelog_date: datetime = datetime.now()
 
+    def changelog_add(self) -> bool:
```
onurctirtir, line 69: maybe:

```suggestion
    def should_add_changelog(self) -> bool:
```

packaging_automation/update_package_properties.py
```diff
@@ -40,22 +66,36 @@ class PackagePropertiesParams:
...
+
     def spec_file_name(self) -> str:
-        return spec_file_name(self.project_name)
+        return spec_file_name(self.supported_project.value.name)
+
+    def pkgvars_template_file_name(self) -> str:
+        return f"{self.supported_project.value.name}-pkgvars.tmpl"
+
+    def rpm_spec_template_file_name(self) -> str:
+        return f"{self.supported_project.value.name}.spec.tmpl"
```
onurctirtir, line 79: I think those three could be `@property` along with many functions of this class, but we can improve that in another pr too.

packaging_automation/update_package_properties.py
```diff
@@ -192,25 +237,25 @@ def update_rpm_spec(package_properties_params: PackagePropertiesParams, spec_ful
...
+                              project_name=package_properties_params.supported_project.value.name,
                               fancy_version_no=package_properties_params.fancy_version_number, changelog=changelog)
     with open(spec_full_path, "w+") as writer:
         writer.write(content)
 
 
 def validate_package_properties_params_for_update_all_changes(package_props: PackagePropertiesParams):
-    if not package_props.project_name:
-        raise ValueError("Project Name should not be empty")
     if package_props.fancy_version_number < 0:
```
onurctirtir, line 259: why are we removing this check ?

packaging_automation/update_package_properties.py
```diff
@@ -226,43 +271,69 @@ def update_all_changes(github_token: non_empty(non_blank(str)), package_properti
...
-    parser.add_argument('--name')
-    parser.add_argument('--date', nargs='?', default=datetime.now())
+    parser.add_argument('--gh_token', required=True)
+    parser.add_argument('--prj_name', choices=[r.name for r in SupportedProjects])
+    parser.add_argument('--tag_name', required=True)
+    parser.add_argument('--fancy_ver_no', type=int, choices=range(1, 10), default=1)
+    parser.add_argument('--email', required=True)
+    parser.add_argument('--name', required=True)
+    parser.add_argument('--date')
+    parser.add_argument("--pipeline", action="store_true")
```
onurctirtir, line 295: 1) What is the usage of `pipeline` param and can we document it in README ?
2) I think we should remove `fancy` from README too

packaging_automation/update_package_properties.py
```diff
@@ -226,43 +271,69 @@ def update_all_changes(github_token: non_empty(non_blank(str)), package_properti
...
     parser.add_argument('--exec_path')
+    parser.add_argument('--is_test', action="store_true")
     arguments = parser.parse_args()
 
-    if not string_utils.is_integer(arguments.fancy_ver_no):
-        raise ValueError(f"fancy_ver_no is expected to be numeric actual value {arguments.fancy_ver_no}")
+    prj_ver = get_project_version_from_tag_name(arguments.tag_name)
 
-    exec_date = datetime.strptime(arguments.date, '%Y.%m.%d %H:%M:%S %z')
+    supported_project = SupportedProjects[arguments.prj_name]
```
onurctirtir, line 302:
```suggestion
    project = SupportedProjects[arguments.prj_name]
```

packaging_automation/update_package_properties.py
```diff
@@ -67,6 +107,13 @@ def debian_trailer(self):
         return f" -- {self.name_surname} <{self.microsoft_email}>  {formatted_date} \n "
 
 
+def get_enum_from_changelog_project_name(changelog_project_name) -> SupportedProjects:
+    for e in SupportedProjects:
+        if e.value.changelog_project_name == changelog_project_name:
+            return e
+    raise ValueError(f"{changelog_project_name} could not be found in supported project changelog project names.")
```
onurctirtir, line 114: minor:

```suggestion
    raise ValueError(f"{changelog_project_name} could not be found in supported project changelog names.")
```

packaging_automation/update_package_properties.py
onurctirtir: can you open an issue for that ?

packaging_automation/upload_to_package_cloud.py
gurkanindibay: It is used for finding the current branch. If current branch is not main branch, packages are not pushed

packaging_automation/templates/citus-enterprise.spec.tmpl
gurkanindibay: Fixed

packaging_automation/update_package_properties.py
gurkanindibay: Fixed

packaging_automation/update_package_properties.py
gurkanindibay: Fixed

packaging_automation/update_package_properties.py
gurkanindibay: Fixed

packaging_automation/update_package_properties.py
gurkanindibay: Actually I couldn't understand can you give me a little bit detail?

packaging_automation/update_package_properties.py
gurkanindibay: Fixed

packaging_automation/update_package_properties.py
gurkanindibay: Because it is validated on main function and there is no need to validate here. 
parser.add_argument('--prj_name', choices=[r.name for r in SupportedProject])

packaging_automation/update_package_properties.py
gurkanindibay: we can execute this script either from our computer or in pipeline. When using pipeline parameter code is not cloned since pipeline clones it beforehand.

gurkanindibay: I'll update Readme for all requests

packaging_automation/update_package_properties.py
gurkanindibay: Fixed

packaging_automation/update_package_properties.py
gurkanindibay: Fixed

packaging_automation/upload_to_package_cloud.py
onurctirtir: To verify if I understood correctly, is `exec_forder_path` to specify the branch we are working on ? 
Or is it another parameter that you explained ?

packaging_automation/update_package_properties.py
onurctirtir: >I'll update Readme for all requests

waiting for the update then

onurctirtir: > we can execute this script either from our computer or in pipeline. When using pipeline parameter code is not cloned since pipeline clones it beforehand.

Is there any reason to not clone the repo always via the script, even when using pipeline on CI.
I think it might be a slightly cleaner that way, what do you think ?

packaging_automation/update_package_properties.py
onurctirtir: In python, it is pretty usual to use `@property` decorator for the member functions that are doing simple calculations that are not modifying object state but returning a "property" of given object without storing that property as a member variable.

I think we should use that decorator for the member functions that are evaluating such strings etc.

https://www.machinelearningplus.com/python/python-property/
(especially the second section)

onurctirtir: But we can do that in another pr and open an issue for that for now.

onurctirtir (commented):
Added some comments to previous ones.

packaging_automation/update_package_properties.py
gurkanindibay: The problem is that pipeline will execute in the packaging repo so code is already there. If we checkout the code here again, we will do an unnecessary job. For example when I execute the pipeline if pipeline is true, exec path will be used which will be ${pwd} most probably.

packaging_automation/update_package_properties.py
gurkanindibay: Fixed

packaging_automation/upload_to_package_cloud.py
gurkanindibay: exactly. We need to know the "packaging" directory we are working on to get the current branch. We will execute this code from pipeline. Otherwise I need to get the current branch from github which is a little bit tricky. But if you want I can get the branch name from github actions. 
Getting the branch name from github actions is  as below
https://stackoverflow.com/questions/58033366/how-to-get-current-branch-within-github-actions/58034787

onurctirtir: If it wouldn't be a big /complex change, I think "github actions approach" looks more clear

packaging_automation/update_package_properties.py
onurctirtir: oh now understood what you mean, resolving the comment then

packaging_automation/upload_to_package_cloud.py
gurkanindibay: Fixed

packaging_automation/upload_to_package_cloud.py
```diff
@@ -116,15 +115,15 @@ def package_exists(package_cloud_token: str, repo_owner: str, repo_name: str, pa
     parser.add_argument('--package_cloud_api_token', required=True)
     parser.add_argument('--repository_name', required=True, choices=supported_repos)
     parser.add_argument('--output_file_path', required=True)
-    parser.add_argument('--exec_folder_path', required=True)
+    parser.add_argument('--current_branch', required=True)
```
onurctirtir, line 118: Now we have --current_branch param but it is not set via github actions job or something, is it what we want ?

packaging_automation/tests/test_citus_package.py
```diff
@@ -16,6 +16,7 @@
 PACKAGING_SOURCE_FOLDER = "packaging_test"
 PACKAGING_EXEC_FOLDER = f"{TEST_BASE_PATH}/{PACKAGING_SOURCE_FOLDER}"
 BASE_OUTPUT_FOLDER = f"{PACKAGING_EXEC_FOLDER}/packages"
+CURRENT_BRANCH = "all-citus"
 MAIN_BRANCH = "all-citus"
```
onurctirtir, line 20: I think we can remove those two global variables and just set them in `test_upload_to_package_cloud` since they don't seem to be referenced from elsewhere

packaging_automation/upload_to_package_cloud.py
gurkanindibay: it will be set on Github Actions pipeline. I will send the pipeline code in another pipeline in packaging repo

packaging_automation/tests/test_citus_package.py
gurkanindibay: Fixed

onurctirtir (approved):

PR #134, opened 2021-08-02
Package cloud statistics  download cron schedule fix

gurkanindibay:
There is a problem in Package Cloud Statistics Download Cron schedule as below. This PR fixes invalid repo names in schedule pipeline and adds a repo name column into statistics table.
https://github.com/citusdata/tools/actions/runs/1088494331

### DIFF
--- a/packaging_automation/package_cloud_statistics_collector.py
+++ b/packaging_automation/package_cloud_statistics_collector.py
@@ -9,7 +9,6 @@ from typing import List, Any
 import requests
 from sqlalchemy import Column, INTEGER, DATE, TIMESTAMP, String
 
-from release.docker.tools.packaging_automation.package_cloud_statistics_collector import PackageCloudRepos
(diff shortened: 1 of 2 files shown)


packaging_automation/package_cloud_statistics_collector.py
```diff
@@ -34,6 +34,7 @@ class PackageCloudDownloadStats(Base):
     __tablename__ = "package_cloud_download_stats"
     id = Column(INTEGER, primary_key=True, autoincrement=True)
     fetch_date = Column(TIMESTAMP, nullable=False)
+    repo = Column(String, nullable=False)
```
halilozanakgul, line 37: We could use `PackageCloudRepos` as data type instead of string.

halilozanakgul (changes requested):

halilozanakgul (approved):

PR #150, opened 2021-08-09
update_package_properties pipeline branch problem

### LINKED ISSUE #149: update_package_properties pipeline execution problem
while update_package properties pipeline is executed in pipeline, only current active branch for the pipeline is checked out. So while executing checkout operation for "all-citus" it fails. Additionally there is no need to checkout all-citus in the code while working in pipeline 
Problem while executing
error: pathspec 'all-citus' did not match any file(s) known to git

gurkanindibay:
update_package_properties script can be executed both manually and in pipeline. While executing in pipeline, all branches are not closed so checking out the main packaging branch (e.g. ell-citus) causes problem in PR branches since all-citus is not fetched at the time of pipeline execution.
Checking out handled if pipeline flag is set and also code deletion is done in manual execution case i.e. when pipeline is false

Fixes #149

### DIFF
--- a/packaging_automation/update_package_properties.py
+++ b/packaging_automation/update_package_properties.py
@@ -349,5 +349,4 @@ if __name__ == "__main__":
         run(f'git push --set-upstream origin {pr_branch}')
         create_pr(arguments.gh_token, pr_branch, commit_message, REPO_OWNER, PROJECT_NAME,
                   project.value.packaging_branch)
-        if not arguments.pipeline:
-            remove_cloned_code(execution_path)
+        remove_cloned_code(execution_path)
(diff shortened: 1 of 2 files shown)


packaging_automation/update_package_properties.py
```diff
@@ -348,4 +349,5 @@ def update_all_changes(github_token: non_empty(non_blank(str)), package_properti
         run(f'git push --set-upstream origin {pr_branch}')
         create_pr(arguments.gh_token, pr_branch, commit_message, REPO_OWNER, PROJECT_NAME,
                   project.value.packaging_branch)
-        remove_cloned_code(execution_path)
+        if not arguments.pipeline:
```
hanefi, line 352: Why do we not remove the cloned code on pipelines? Any particular reason?

gurkanindibay: because pipeline machines are disposable. So there is no need to delete the code there since machine is disposed at the end of the pipeline execution

hanefi: If an operation does not take too much time to complete, we can skip the conditionals and perform it everywhere. I think this hurts readability and we should revert this change here.

gurkanindibay: Fixed

hanefi (approved):

PR #159, opened 2021-08-10
Add default branch check for nightly image push

gurkanindibay:


hanefi:
The code changes seem ok, but I have some issues with the high level design.

I think we should let the decision on when to push docker images in the GitHub actions definitions. The python code should have separate interfaces to (1) build and to (2) publish docker images. It should be the responsibility of the workflow definition to decide on if only (1) should happen, or both (1) and (2).

Does that make sense?

gurkanindibay:
> The code changes seem ok, but I have some issues with the high level design.
> 
> I think we should let the decision on when to push docker images in the GitHub actions definitions. The python code should have separate interfaces to (1) build and to (2) publish docker images. It should be the responsibility of the workflow definition to decide on if only (1) should happen, or both (1) and (2).
> 
> Does that make sense?

Actually there are two operations
1. Build
2. Build and publish
In pipeline only publish does not makes sense since virtual machines are disposable and does not store docker builds.
Therefore since Build and publish is the superset of build there is no benefit of seperating build and publish on both user perspective and pipeline perspective
On the other hand when I look at readability perspective, it makes sense. But I need to split the script, change unit tests etc. Utku and Metin are putting pressure on me to finish these scripts. If it is crucial, I can open an issue for this and prioritize this to be developed in a more suitable timeline

hanefi (approved):
Please create an issue on why we need to have the logic on what packages to push on a `.yml` file on docker repo, and not inside the python scripts.

I am ok with merging this one as is, just to unblock you. I do not agree that this is feasible, and it can make maintaining this machinery hard for us in the future.

PR #175, opened 2021-09-01
Prospector static code analysis tool support addition

gurkanindibay:
There is no atatic code analysis tool integrated into tools repo code. It makes reviewing codes hard. I used prospector static code analysis tool which is a container for multiple code analysis tool.
For detailed info you can check http://prospector.landscape.io/en/master/
If you want to locally check the code first you can install prospector using 
```pip install prospector[with_everything]```
then you execute prospector in tools repo by just executing 
```prospector```

When I first started, I get an ouput like attached and fixed all of the outputs
Pipeline is integra

[prospector.txt](https://github.com/citusdata/tools/files/7103783/prospector.txt)

prospector is integrated into tool-tests pipeline as below
https://github.com/citusdata/tools/blob/a66910e3a1ed15e465eecadbcf7a7e3ccf868ee9/.github/workflows/tool-tests.yml#L42-L43

Prospector configuration file is as below.
https://github.com/citusdata/tools/blob/static_code/.prospector.yaml

### CODE REVIEW

File: packaging_automation/common_tool_methods.py [hunk 1]
```diff
@@ -113,8 +119,10 @@ def run(command, *args, **kwargs):
 
 
 def run_with_output(command, *args, **kwargs):
-    result = subprocess.run(shlex.split(command), *args, capture_output=True,
-                            **kwargs)
+    # this method's main objective is to return output. Therefore it is caller's responsibility to handle
```
onurctirtir, line 122: thank you for commenting pylint ignores!

packaging_automation/update_package_properties.py
```diff
@@ -278,6 +283,7 @@ def validate_package_properties_params_for_update_all_changes(package_props: Pac
 
 
 @validate_parameters
+# pylint: disable=no-value-for-parameter
```
onurctirtir, line 286: maybe comment the reasoning for that ignore rule here, as we did in others

.prospector.yaml
```diff
@@ -0,0 +1,17 @@
+test-warnings: true
+doc-warnings: false
+member-warnings: false
+
+ignore-paths:
+  - docs
+ignore-patterns:
+  - (^|/)skip(this)?(/|$)
+autodetect: true
+#max-line-length: 88
```
onurctirtir, line 10: enabling this could be nice, but leaving the decision to you

.prospector.yaml
```diff
@@ -0,0 +1,17 @@
+test-warnings: true
+doc-warnings: false
+member-warnings: false
+
+ignore-paths:
+  - docs
+ignore-patterns:
+  - (^|/)skip(this)?(/|$)
```
onurctirtir, line 8: what is that for ?

.prospector.yaml
```diff
@@ -0,0 +1,17 @@
...
+autodetect: true
+#max-line-length: 88
+
+pylint:
+  disable:
+    # parameter_validations library methods give this error
+    - no-value-for-parameter
```
onurctirtir, line 15: can we ignore this only in the places where we need to do so ?

.prospector.yaml
```diff
@@ -0,0 +1,17 @@
+test-warnings: true
```
onurctirtir, line 1: can we, in general, comment the configs in this file when necessary ?

.prospector.yaml
```diff
@@ -0,0 +1,17 @@
...
+  - (^|/)skip(this)?(/|$)
+autodetect: true
+#max-line-length: 88
+
+pylint:
+  disable:
+    # parameter_validations library methods give this error
+    - no-value-for-parameter
+  options:
+    max-args: 6
```
onurctirtir, line 17: comment ?

onurctirtir (approved):
Left some minor comments, thank you for enabling that!

.prospector.yaml
gurkanindibay: Yes we can add comments if we want

.prospector.yaml
gurkanindibay: default was 5 I think 6 will be more suitable. Do you mean why I made it 6? 5 is most of the times not enough. It is hard for me to make methods with less than 6 params is very difficult. In the future, we can evaluate by the way

.prospector.yaml
gurkanindibay: Yes I removed the general rule for no-value-for-parameter and added it inline wherever it is needed with comments

packaging_automation/update_package_properties.py
gurkanindibay: Fixed

PR #177, opened 2021-09-06
Adds packaging images release name support into citus package

gurkanindibay:
Packaging images creation release names are different than standard release names. Packaging images release names are seperated by comma but our general release names are seperated by '/'. Some release names are totally different as well like redhat and oracle. 
In this PR, I added a map to support the ones seperated with comma
This PR is related with the one below
https://github.com/citusdata/packaging/pull/745

Additionally, there was problems in pgxn and ol/8 tests.I fixed these problems in this PR as well


hanefi (commented):
What do you think about using only `/` characters and stop using `,` to separate OS name and release identifiers? Do you think it would be easier to handle?

packaging_automation/tests/test_citus_package.py
```diff
@@ -22,37 +22,40 @@
...
+            f"git clone --branch {packaging_branch_name} https://github.com/citusdata/packaging.git"
+            f" {PACKAGING_EXEC_FOLDER}")
 
 
-def teardown_module():
-    if os.path.exists("packaging_test"):
-        run("rm -r packaging_test")
+# def teardown_module():
+#     if os.path.exists("packaging_test"):
+#         run("rm -r packaging_test")
```
hanefi, line 55: I do not like these commented out lines

packaging_automation/citus_package.py
```diff
@@ -71,16 +85,21 @@ class PostgresVersionDockerImageType(Enum):
...
     if len(parts) == 1 and parts[0] == "pgxn":
         os_name = "pgxn"
         os_release = ""
     else:
         os_name = parts[0]
         os_release = parts[1]
+        # When building packaging images oraclelinux and centos is used instead of el and ol
+        # Below two lines handles conversion betwwen them
+        os_name = os_name if os_name != "oraclelinux" else "ol"
+        os_name = os_name if os_name != "centos" else "el"
```
hanefi, line 102:
```suggestion
        os_name = "ol" if os_name == "oraclelinux" else os_name
        os_name = "el" if os_name == "centos" else os_name
```

I think this one is easier to read and understand.

hanefi (changes requested):

gurkanindibay:
> What do you think about using only `/` characters and stop using `,` to separate OS name and release identifiers? Do you think it would be easier to handle?

there are lots of place to change and it is risky and will take too much time. I need to finish conversion ASAP

packaging_automation/tests/test_citus_package.py
gurkanindibay: My misatake. Uncommented lines

hanefi (approved):
I am not totally comfortable with adding yet another argument to this cli tool. The arguments are not documented well and it is not so easy to understand their purposes without digging deeper into the codebase. I have a hard time creating packages on my local machine now, and it appears that it will be even harder than before.

I hope that we can make these tools more user friendly in the future.

PR #181, opened 2021-09-15
Fix for debian based builds output validation

gurkanindibay:
While applying output_validation, I found out that lintian systems have also warnings including E:. I added this pattern and changed unit tests


packaging_automation/packaging_warning_handler.py
```diff
@@ -40,33 +40,38 @@ def validate_output(output: str, ignore_file_path: str, package_type: PackageTyp
...
     is_deb_warning_line = False
     is_rpm_warning_line = False
     for output_line in output_lines:
+
         if package_type == PackageType.deb:
-            if "warning" in output_line.lower() or is_deb_warning_line:
-                if "Warnings generated by release docker build" in output_line:
-                    is_deb_warning_line = True
-                    continue
+            if debian_lintian_starter in output_line:
```
gurkanindibay, line 55: Here there were some error preventing debian warnings to be handled. First lintian starter line to be checked to handle W E like errors (like RPM below)

packaging_automation/packaging_warning_handler.py
```diff
@@ -40,33 +40,38 @@ def validate_output(output: str, ignore_file_path: str, package_type: PackageTyp
 
 
 def filter_warning_lines(output_lines: List[str], package_type: PackageType) -> Tuple[List[str], List[str]]:
+    rpm_warning_summary = r"\d+ packages and \d+ specfiles checked; \d+ errors, \d+ warnings."
```
gurkanindibay, line 43: parameterized hardcoded variables

packaging_automation/packaging_warning_handler.py
```diff
@@ -40,33 +40,38 @@ def validate_output(output: str, ignore_file_path: str, package_type: PackageTyp
...
+    rpm_lintian_starter = 'Executing "/usr/bin/rpmlint -f /rpmlintrc'
+    debian_lintian_starter = "Now running lintian"
+    lintian_warning_error_pattern = r".*: [W|E]: .*"
```
gurkanindibay, line 46: THis is the regex that handles E based lines

packaging_automation/tests/files/packaging_warning/sample_warning_build_output_deb.txt
```diff
@@ -129,13 +129,15 @@ gpg: Total number processed: 1
 gpg:               imported: 1  (RSA: 1)
 gpg:       secret keys read: 1
 gpg:   secret keys imported: 1
-Warnings generated by release docker build :
```
hanefi, line 132: I would not want to lose this line here

gurkanindibay: I need to because this is not the starter line for lintian checks any more you can check it below
https://github.com/citusdata/packaging/runs/3619678886?check_suite_focus=true#step:6:3035

hanefi (approved):

PR #184, opened 2021-09-20
Fixes current branch check problem when tag is pushed to push docker tags

gurkanindibay:
After building main docker images, we need to create tagged docker images such as 10.0-alpine by creating new tag on master branch
While tagging, Github Actions clones the code in detached state and when in this state we can not get current branch . Since we are using current branch to decide whether images weill published or not, and we can not use it in tag push case , we need to put tag existence case on master branch to be able to handle tag push case
Som improvements are suggested. These are in the issue below
https://github.com/citusdata/tools/issues/192

### DIFF
--- a/packaging_automation/common_tool_methods.py
+++ b/packaging_automation/common_tool_methods.py
@@ -1,19 +1,17 @@
-import base64
 import os
 import re
 import shlex
 import subprocess
 from datetime import datetime
 from enum import Enum
-from typing import Dict, List
-from typing import Tuple
+from typing import Dict, List,Tuple
 
+import git
 import gnupg
 import pathlib2
 import requests
-from git import Repo, GitCommandError
-import git
-from github import Repository, PullRequest, Commit, Github
+from git import GitCommandError,Repo
+from github import Commit, Github, PullRequest,Repository
 from jinja2 import Environment, FileSystemLoader
 from parameters_validation import validate_parameters
 
--- a/packaging_automation/common_tool_methods.py
+++ b/packaging_automation/common_tool_methods.py
@@ -11,7 +11,8 @@ from typing import Tuple
 import gnupg
 import pathlib2
 import requests
-from git import Repo
+from git import Repo,GitCommandError
+import git
 from github import Repository, PullRequest, Commit, Github
 from jinja2 import Environment, FileSystemLoader
 from parameters_validation import validate_parameters
@@ -299,10 +300,29 @@ def prepend_line_in_file(file: str, match_regex: str, append_str: str) -> bool:
 
     return has_match
 
+def is_tag_on_branch(tag_name: str, branch_name: str):
+    g = git.Git(os.getcwd())
+    try:
+        branches_str = g.execute(["git", "branch", "--contains", f"tags/{tag_name}"])
+        branches = remove_prefix(branches_str, "*").split("\n")
+        if len(branches) > 0:
+            for branch in branches:
+                if branch.strip() == branch_name:
+                    return True
+        return False
+    except GitCommandError:
+        return False
+
+
+
 
 def get_current_branch(working_dir: str) -> str:
     repo = get_new_repo(working_dir)
-    return repo.active_branch.name
+    try:
+        branch_name = repo.active_branch.name
+    except TypeError:
+        branch_name = 'Detached'
+    return branch_name
 
 
 def remote_branch_exists(branch_name: str, working_dir: str) -> bool:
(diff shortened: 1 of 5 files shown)


packaging_automation/common_tool_methods.py
```diff
@@ -299,10 +300,29 @@ def prepend_line_in_file(file: str, match_regex: str, append_str: str) -> bool:
 
     return has_match
 
+def is_tag_on_branch(tag_name: str, branch_name: str):
```
gurkanindibay, line 303: This is the new method to check whether branch contains tag

packaging_automation/publish_docker.py
```diff
@@ -147,8 +147,9 @@ def publish_tagged_docker_images(docker_image_type, tag_name: str, current_branc
     docker_client.images.build(dockerfile=docker_image_info_dict[docker_image_type]['file-name'],
                                tag=docker_image_name,
                                path=".")
-    if current_branch != DEFAULT_BRANCH_NAME:
-        print(f"Since current branch {current_branch} is not equal to {DEFAULT_BRANCH_NAME} tags will not be pushed.")
+    if is_tag_on_branch(tag_name=tag_name, branch_name=DEFAULT_BRANCH_NAME):
```
gurkanindibay, line 150: In case of tag push, tag contains check is used instead of branch name

packaging_automation/publish_docker.py
```diff
@@ -216,23 +219,42 @@ def validate_and_extract_manual_exec_params(manual_trigger_type_param: str, tag_
     return manual_trigger_type_param
 
 
+def get_image_publish_status(github_ref: str, is_test: bool):
```
gurkanindibay, line 222: Centralized the image publish status to manage the publish status in one place

packaging_automation/tests/test_common_tool_methods.py
```diff
@@ -16,7 +16,7 @@
     local_branch_exists, get_last_commit_message, get_prs_for_patch_release, filter_prs_by_label, process_template_file,
     remove_prefix, delete_all_gpg_keys_by_name, define_rpm_public_key_to_machine,
     delete_rpm_key_by_name, get_gpg_fingerprints_by_name, run_with_output, rpm_key_matches_summary,
-    DEFAULT_ENCODING_FOR_FILE_HANDLING, DEFAULT_UNICODE_ERROR_HANDLER)
+    DEFAULT_ENCODING_FOR_FILE_HANDLING, DEFAULT_UNICODE_ERROR_HANDLER, is_tag_on_branch)
```
hanefi, line 19: organize imports please.

My suggestion:
```python
from ..common_tool_methods import (DEFAULT_ENCODING_FOR_FILE_HANDLING,
                                   DEFAULT_UNICODE_ERROR_HANDLER,
                                   append_line_in_file,
                                   define_rpm_public_key_to_machine,
                                   delete_all_gpg_keys_by_name,
                                   delete_rpm_key_by_name, filter_prs_by_label,
                                   find_nth_matching_line_and_line_number,
                                   find_nth_occurrence_position,
                                   get_current_branch,
                                   get_gpg_fingerprints_by_name,
                                   get_last_commit_message, get_minor_version,
                                   get_patch_version_regex,
                                   get_project_version_from_tag_name,
                                   get_prs_for_patch_release,
                                   get_upcoming_minor_version,
                                   get_version_details, is_major_release,
                                   is_tag_on_branch, local_branch_exists,
                                   prepend_line_in_file, process_template_file,
                                   remote_branch_exists, remove_prefix,
                                   remove_text_with_parenthesis,
                                   replace_line_in_file,
                                   rpm_key_matches_summary, run,
                                   run_with_output, str_array_to_str)
from .test_utils import generate_new_gpg_key
```

gurkanindibay: Fixed as you wish

hanefi:
@gurkanindibay I do not really get why we need to have a really convoluted solution to this problem.

I think we have 2 scenarios where we push images to DockerHub.

1.
```yaml
on:
  push:
    branches:
      - master
```

2.
```yaml
on:
  push:
    tags:
      - "**"
```

In either case, we should be able to get the ref properly by using `${GITHUB_REF#refs/*/}` that will resolve to the branch name in case 1, and tag name in case 2.

Note that you may need separate workflow definitions to allow this.

Let's have a chat offline and create a better design for solving this problem.

hanefi (approved):

PR #186, opened 2021-09-21
Adds test and pipeline check to delete resources

gurkanindibay:
Without this check resources are deleted an pipelines get errors while trying to delete the checked out resource
https://github.com/citusdata/packaging/runs/3555246468?check_suite_focus=true#step:14:2

### DIFF
--- a/packaging_automation/update_package_properties.py
+++ b/packaging_automation/update_package_properties.py
@@ -354,3 +354,5 @@ if __name__ == "__main__":
         create_pr(arguments.gh_token, pr_branch, commit_message, REPO_OWNER, PROJECT_NAME,
                   project.value.packaging_branch)
         remove_cloned_code(execution_path)
+    if not arguments.is_test and not arguments.pipeline:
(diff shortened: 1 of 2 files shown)


packaging_automation/update_package_properties.py
```diff
@@ -354,3 +354,5 @@ def update_all_changes(github_token: non_empty(non_blank(str)), package_properti
         create_pr(arguments.gh_token, pr_branch, commit_message, REPO_OWNER, PROJECT_NAME,
                   project.value.packaging_branch)
         remove_cloned_code(execution_path)
+    if not arguments.is_test and not arguments.pipeline:
+        remove_cloned_code(execution_path)
```
onurctirtir, line 358: we already remove this directory in line:356 ? why do we need this change ?

onurctirtir (approved):

packaging_automation/update_package_properties.py
gurkanindibay: I will remove from 356

PR #197, opened 2021-10-15
Fixes postgres 14 changes for update_docker

gurkanindibay:
- [x] First pkgvars addition PR, https://github.com/citusdata/docker/pull/281  should be merged then unit tests depends on the branch should be changed into master

### DIFF
--- a/packaging_automation/templates/docker/postgres-13/postgres-13.tmpl.dockerfile
+++ b/packaging_automation/templates/docker/postgres-13/postgres-13.tmpl.dockerfile
@@ -1,7 +1,6 @@
-
 # This file is auto generated from it's template,
-# see citusdata/tools/packaging_automation/templates/docker/postgres-14/postgres-14.tmpl.dockerfile.
-FROM postgres:14rc1
+# see citusdata/tools/packaging_automation/templates/docker/latest/latest.tmpl.dockerfile.
+FROM postgres:13.4
 ARG VERSION={{project_version}}
 LABEL maintainer="Citus Data https://citusdata.com" \
       org.label-schema.name="Citus" \
@@ -20,7 +19,7 @@ RUN apt-get update \
        ca-certificates \
        curl \
     && curl -s https://install.citusdata.com/community/deb.sh | bash \
-    && apt-get install -y postgresql-$PG_MAJOR-citus-{{project_minor_version}}=$CITUS_VERSION \
+    && apt-get install -y postgresql-$PG_MAJOR-citus-{{project_minor_version}}.=$CITUS_VERSION \
                           postgresql-$PG_MAJOR-hll=2.16.citus-1 \
                           postgresql-$PG_MAJOR-topn=2.4.0 \
     && apt-get purge -y --auto-remove curl \
--- a/packaging_automation/templates/docker/postgres-13/postgres-13.tmpl.dockerfile
+++ b/packaging_automation/templates/docker/postgres-13/postgres-13.tmpl.dockerfile
@@ -1,5 +1,5 @@
 # This file is auto generated from it's template,
-# see citusdata/tools/packaging_automation/templates/docker/latest/latest.tmpl.dockerfile.
+# see citusdata/tools/packaging_automation/templates/docker/postgres-13/postgres-13.tmpl.dockerfile.
 FROM postgres:{{postgres_version}}
 ARG VERSION={{project_version}}
 LABEL maintainer="Citus Data https://citusdata.com" \
@@ -19,7 +19,7 @@ RUN apt-get update \
        ca-certificates \
        curl \
     && curl -s https://install.citusdata.com/community/deb.sh | bash \
-    && apt-get install -y postgresql-$PG_MAJOR-citus-{{project_minor_version}}.=$CITUS_VERSION \
+    && apt-get install -y postgresql-$PG_MAJOR-citus-{{project_minor_version}}=$CITUS_VERSION \
                           postgresql-$PG_MAJOR-hll=2.16.citus-1 \
                           postgresql-$PG_MAJOR-topn=2.4.0 \
     && apt-get purge -y --auto-remove curl \
--- a/packaging_automation/templates/docker/postgres-13/postgres-13.tmpl.dockerfile
+++ b/packaging_automation/templates/docker/postgres-13/postgres-13.tmpl.dockerfile
@@ -1,6 +1,6 @@
 # This file is auto generated from it's template,
 # see citusdata/tools/packaging_automation/templates/docker/latest/latest.tmpl.dockerfile.
-FROM postgres:13.4
+FROM postgres:13.5
 ARG VERSION={{project_version}}
 LABEL maintainer="Citus Data https://citusdata.com" \
       org.label-schema.name="Citus" \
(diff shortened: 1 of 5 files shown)


packaging_automation/tests/test_update_docker.py
```diff
@@ -21,7 +24,8 @@
 
 def setup_module():
     if not os.path.exists("docker"):
-        run("git clone https://github.com/citusdata/docker.git")
+        # TODO Remove branch name after merge of docker/PR# 281
```
gurkanindibay, line 27: Here branch name will be removed after merging https://github.com/citusdata/docker/pull/281

onurctirtir:
Will try to review this pr today.

packaging_automation/templates/docker/postgres-13/postgres-13.tmpl.dockerfile
```diff
@@ -1,7 +1,6 @@
-
 # This file is auto generated from it's template,
-# see citusdata/tools/packaging_automation/templates/docker/postgres-14/postgres-14.tmpl.dockerfile.
-FROM postgres:14rc1
+# see citusdata/tools/packaging_automation/templates/docker/latest/latest.tmpl.dockerfile.
```
hanefi, line 2: PG 13 is not the latest major. I think there is a mistake in this comment here.

packaging_automation/templates/docker/postgres-13/postgres-13.tmpl.dockerfile
```diff
@@ -20,7 +19,7 @@ RUN apt-get update \
        ca-certificates \
        curl \
     && curl -s https://install.citusdata.com/community/deb.sh | bash \
-    && apt-get install -y postgresql-$PG_MAJOR-citus-{{project_minor_version}}=$CITUS_VERSION \
+    && apt-get install -y postgresql-$PG_MAJOR-citus-{{project_minor_version}}.=$CITUS_VERSION \
```
hanefi, line 22: Why do we have the extra `.` character here?

packaging_automation/update_docker.py
```diff
@@ -59,17 +61,17 @@ def update_docker_file_alpine(project_version: str, template_path: str, exec_pat
...
-def update_docker_file_for_postgres14(project_version: str, template_path: str, exec_path: str):
+def update_docker_file_for_postgres13(project_version: str, template_path: str, exec_path: str, postgres_version: str):
     content = process_template_file(project_version, template_path,
-                                    docker_templates[SupportedDockerImages.postgres14])
-    dest_file_name = f"{exec_path}/{docker_outputs[SupportedDockerImages.postgres14]}"
+                                    docker_templates[SupportedDockerImages.postgres13], postgres_version)
+    dest_file_name = f"{exec_path}/{docker_outputs[SupportedDockerImages.postgres13]}"
     dir_name = os.path.dirname(dest_file_name)
     if not os.path.exists(dir_name):
         os.makedirs(dir_name)
```
hanefi, line 77: Why do we create missing directories only for pg13 and not pg12?

Can you move the common logic in a function, and reduce repetition here?

packaging_automation/update_docker.py
```diff
@@ -103,48 +105,29 @@ def update_changelog(project_version: str, exec_path: str, postgres_version: str
...
-    update_docker_file_for_postgres14(project_version, template_path, exec_path)
-    update_changelog(project_version, exec_path, postgres_version)
+    update_docker_file_alpine(project_version, template_path, exec_path, latest_postgres_version)
+    update_docker_file_for_postgres12(project_version, template_path, exec_path, postgres_12_version)
+    update_docker_file_for_postgres13(project_version, template_path, exec_path, postgres_13_version)
+    update_changelog(project_version, exec_path, latest_postgres_version)
 
 
-def read_postgres_version(pkgvars_file: str) -> str:
+def read_postgres_version(pkgvars_file: str) -> Tuple[str, str, str]:
```
hanefi, line 124: I think you should change the name of this function now. It returns three versions, not a single one.

packaging_automation/tests/test_update_docker.py
```diff
@@ -5,14 +5,17 @@
...
+                             update_docker_file_alpine, update_docker_file_for_postgres12,
+                             update_docker_file_for_postgres13, update_changelog)
 
 BASE_PATH = os.getenv("BASE_PATH", default=pathlib2.Path(__file__).parents[2])
 TEST_BASE_PATH = f"{BASE_PATH}/docker"
 PROJECT_VERSION = "10.0.3"
-POSTGRES_VERSION = "14.0"
+POSTGRES_14_VERSION = "14.1"
+POSTGRES_13_VERSION = "13.5"
+POSTGRES_12_VERSION = "12.9"
```
hanefi, line 18: Do you plan to update all these tests every time a new PG version is released?

I do not think that it is wise to create a test that needs to updated 4 times a year. Can you make it so that the tests also read the versions from `pkgvars`?

packaging_automation/templates/docker/postgres-13/postgres-13.tmpl.dockerfile
hanefi: And more importantly, how is it possible that the tests passed even though this `apt-install` line seems to be broken?

packaging_automation/templates/docker/postgres-13/postgres-13.tmpl.dockerfile
onurctirtir: +1, I'm also wondering why.

packaging_automation/tests/test_update_docker.py
gurkanindibay: Tests are not for testing that specific version. They are just for testing the code so there is no need to update everytime postgres version changes

packaging_automation/update_docker.py
gurkanindibay: Fixed as you suggested

packaging_automation/update_docker.py
gurkanindibay: Fixed as you suggested

packaging_automation/templates/docker/postgres-13/postgres-13.tmpl.dockerfile
gurkanindibay: Both with dot and without dot works. I added it just to make it same with other files. I removed all of the dots now in all applicable template files

packaging_automation/templates/docker/postgres-13/postgres-13.tmpl.dockerfile
gurkanindibay: Fixed the comment. Previously pg13 template was the latest so the comments remained as it was before.

packaging_automation/update_docker.py
```diff
@@ -103,48 +110,29 @@ def update_changelog(project_version: str, exec_path: str, postgres_version: str
...
-            if not postgres_version:
-                raise ValueError("pkgvars file should include a line with key latest_postgres_version")
-    else:
-        # Setting it because pkgvars does not exist initially
-        postgres_version = "13.4"
-    return postgres_version
+        config = dotenv_values(pkgvars_file)
+        return config["postgres_14_version"], config["postgres_13_version"], config["postgres_12_version"]
 
+    return "14.1", "13.5", "12.9"
```
onurctirtir, line 134: can move those into some constant variables in this file ?

```python
DOCKER_DEFAULT_LATEST_VER = "14.1"
DOCKER_DEFAULT_PG13_VER = "13.5"
DOCKER_DEFAULT_PG12_VER = "12.9"
```

onurctirtir: We could even define them in a brand new python file, then use them both here and also in `test_update_docker.py`.

onurctirtir (approved):
Please address my final comment before merging, thanks.

PR #198, opened 2021-10-20
Adds package test workflow trigger 

gurkanindibay:


.github/workflows/package-tests.yml
```diff
@@ -64,8 +71,15 @@ jobs:
...
-          [ -z ${PROJECT_VERSION} ]&& export PROJECT_VERSION=10.2.1
+          export PROJECT_VERSION_FROM_PACKAGING="${{ needs.metadata.outputs.citus_version }}"
+          if [[ ! -z ${PROJECT_VERSION_FROM_PACKAGING} ]]
+          then
+            export PROJECT_VERSION=${PROJECT_VERSION_FROM_PACKAGING}
+          else
+            export PROJECT_VERSION="${{ github.event.inputs.prj_ver }}"
+          fi
+          echo "Citus Version: ${PROJECT_VERSION} "
+          [ -z ${PROJECT_VERSION} ] && export PROJECT_VERSION=10.2.1
```
onurctirtir, line 82: why do we need this ?

.github/workflows/package-tests.yml
```diff
@@ -64,8 +71,15 @@ jobs:
...
         run: |
-          export PROJECT_VERSION="${{ github.event.inputs.name }}"
-          [ -z ${PROJECT_VERSION} ]&& export PROJECT_VERSION=10.2.1
+          export PROJECT_VERSION_FROM_PACKAGING="${{ needs.metadata.outputs.citus_version }}"
+          if [[ ! -z ${PROJECT_VERSION_FROM_PACKAGING} ]]
+          then
+            export PROJECT_VERSION=${PROJECT_VERSION_FROM_PACKAGING}
+          else
+            export PROJECT_VERSION="${{ github.event.inputs.prj_ver }}"
+          fi
```
onurctirtir, line 80: can you add some comments to explain in which cases we can (and should) read this from $PROJECT_VERSION_FROM_PACKAGING, and when from ${{ github.event.inputs.prj_ver }} ?

.github/workflows/package-tests.yml
onurctirtir: I meant, at what circumstances "PROJECT_VERSION" could be empty ? I think this shouldn't happen ever.

gurkanindibay: When the pipeline is triggered directly on push, project version is empty. in that case for the sake of testing I want the pipeline to be executed to make sure that test pipeline is intact

.github/workflows/package-tests.yml
gurkanindibay: Removed it with removing the usage of plugin

PR #201, opened 2021-10-31
Adds ol/8 support for package cloud

gurkanindibay:


packaging_automation/citus_package.py
```diff
@@ -300,11 +300,18 @@ def get_package_version_from_pkgvars(input_files_dir: str):
     pkgvars_config = dotenv_values(f"{input_files_dir}/pkgvars")
     package_version_with_suffix = pkgvars_config["pkglatest"]
     version_parts = package_version_with_suffix.split(".")
+    # hll is working with minor release format
+    minor_release = False
```
onurctirtir, line 304: couldn't quite understand what you meant with this, could you explain the reason behind this change a bit more ?

gurkanindibay: https://github.com/citusdata/packaging/blob/7dd542aed9093b98bcd6a6aab142e9c483816cb9/pkgvars#L4
When you look at hll version format, it does not have patch release. Since I got error while parsing version, I used such flag

onurctirtir: Then comment sounds incorrect here, maybe just do something like, using the `pkgname` field from `pkgvars` :

```suggestion
has_minor_release_identifier = False if pkgname == 'hll' else True
```

onurctirtir (approved):
please address the comment before merging, thanks

packaging_automation/citus_package.py
gurkanindibay: Fixed

PR #205, opened 2021-11-09
Fixes invalid parameter fetch problem

gurkanindibay:
prj_ver parameter is the input parameter but I was using the wrong parameter. Fixed that parameter

### DIFF
--- a/packaging_automation/test_citus_package.py
+++ b/packaging_automation/test_citus_package.py
@@ -56,17 +56,17 @@ def get_postgres_versions_from_matrix_file(project_version: str) -> List[str]:
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
-    parser.add_argument('--prj_ver', required=True)
+    parser.add_argument('--project_version', required=True)
     parser.add_argument('--pg_major_version')
     parser.add_argument("--os_release", choices=[t.value["name"] for t in TestPlatform])
 
     args = parser.parse_args()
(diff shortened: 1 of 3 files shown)


hanefi (changes requested):
How about renaming `prj_ver` to `project_version` that seems better to read?

I suggest we do the renaming in `workflow_dispatch.inputs.prj_ver`

packaging_automation/test_citus_package.py
```diff
@@ -85,8 +85,8 @@ def get_postgres_versions_from_matrix_file(project_version: str) -> List[str]:
         docker_image_name = f"test:{test_platform.value['docker_image_name']}-{postgres_version}"
         build_command = (f"docker build -t {docker_image_name} "
                          f"-f {test_platform.value['docker_image_name']}/Dockerfile "
-                         f"--build-arg CITUS_VERSION={args.prj_ver} --build-arg PG_MAJOR={postgres_version}   "
-                         f"--build-arg CITUS_MAJOR_VERSION={minor_prj_ver} .")
+                         f"--build-arg CITUS_VERSION={args.project_version} --build-arg PG_MAJOR={postgres_version}   "
```
hanefi, line 88:
```suggestion
                         f"--build-arg CITUS_VERSION={args.project_version} --build-arg PG_MAJOR={postgres_version}"
```

hanefi (approved):
LGTM, with some minor comments on extra spaces that are not really necessary

PR #207, opened 2021-11-18
Remove deprecated perl functions for packaging

hanefi:
Remove deprecated perl scripts and all the references to them﻿


onurctirtir:
I guess we still don't have the modernized version of upgrade_changelog, so we should not remove that one.

gurkanindibay:
> I guess we still don't have the modernized version of prepare_changelog, so we should not remove that one.

Since we can not remove prepare_changelog we can not remove automated_packaging as well. So automated_packaging mentions should be removed as well

PR #211, opened 2021-11-30
Fixes tag check on branch code

gurkanindibay:
Unit tests was failing for tag check operation. There was a bug on stripping '*' from branch name. Fixed the issue. Previously, I changed the logic but after the comments and detailed investiogation I found out that there is no need for a logic change so I rolledback and applied the solution that keeps the same approach but fixed the bug.

Problem here is current branch starts with * character and previously I strip the fist character of the unsplitted text. However, order of the current branch may change and may be somewhere in between unsplitted text so I splitted the text and removed * character for each branch

### DIFF
--- a/packaging_automation/common_tool_methods.py
+++ b/packaging_automation/common_tool_methods.py
@@ -322,11 +322,11 @@ def is_tag_on_branch(tag_name: str, branch_name: str):
     g = git.Git(os.getcwd())
     try:
         branches_str = g.execute(["git", "branch", "--contains", f"tags/{tag_name}"])
-        branches = remove_prefix(branches_str, "*").split("\n")
+        branches = branches_str.split("\n")
         print("Branches str:" + branches_str)
         if len(branches) > 0:
             for branch in branches:
-                if branch.strip() == branch_name:
+                if remove_prefix(branch, "*").strip() == branch_name:
                     return True
         return False
     except GitCommandError as e:
--- a/packaging_automation/tests/test_common_tool_methods.py
+++ b/packaging_automation/tests/test_common_tool_methods.py
@@ -44,6 +44,7 @@ GITHUB_TOKEN = os.getenv("GH_TOKEN")
 BASE_PATH = pathlib2.Path(__file__).parents[1]
 TEST_BASE_PATH = pathlib2.Path(__file__).parent.absolute()
 TEST_GPG_KEY_NAME = "Citus Data <packaging@citusdata.com>"
+UNIT_TEST_BRANCH = "unit_test_branch"
 
 
 def test_find_nth_occurrence_position():
@@ -84,7 +85,10 @@ def test_get_version_details():
 
 def test_is_tag_on_branch():
     current_branch = get_current_branch(os.getcwd())
-    assert is_tag_on_branch("v0.8.3", current_branch)
+    run(f"git checkout develop")
+    assert is_tag_on_branch("v0.8.3", "develop")
+    assert not is_tag_on_branch("v1.8.3", "develop")
+    run(f"git checkout {current_branch}")
 
 
 def test_replace_line_in_file():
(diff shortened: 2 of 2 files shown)


hanefi:
Can you provide some links to PRs, branch links when applicable?

> Unit tests was failing for tag check operation.

Please share a link to a failing check. And possibly share the errors in the failing test in here as well.

> I changed the logic and tested it.

Where are these changes? Do you have a PR?

> Tesing using develop branch is problematic since the new code does not exist on develop so I used a new unit test branch to be able to test easily. I made this new branch protected to prevent deletion

What is the name of this branch? Can you share a link?

gurkanindibay:
> https://github.com/citusdata/tools/runs/4363804152?check_suite_focus=true
> > Can you provide some links to PRs, branch links when applicable?
> > 
> > > Unit tests was failing for tag check operation.
> > 
> > Please share a link to a failing check. And possibly share the errors in the failing test in here as well.
> > 
> > > I changed the logic and tested it.
> > 
> > Where are these changes? Do you have a PR?
> > 
> >  > Tesing using develop branch is problematic since the new code does not exist on develop so I used a new unit test branch to be able to test easily. I made this new branch protected to prevent deletion
> > 
> > What is the name of this branch? Can you share a link?

packaging_automation/common_tool_methods.py
```diff
@@ -321,19 +321,20 @@ def prepend_line_in_file(file: str, match_regex: str, append_str: str) -> bool:
...
-        print("Branches str:" + branches_str)
-        if len(branches) > 0:
-            for branch in branches:
-                if branch.strip() == branch_name:
+        tags = g.execute(["git", "tag", "--merged", f"{branch_name}"])
+        tags = tags.split()
+        print("Tags:", tags)
+        if len(tags) > 0:
+            for tag in tags:
+                if tag.strip() == tag_name:
```
onurctirtir, line 329: Effectively, `git branch --contains` & `git tag --merged` should both work to check if given tag is on a certain branch.

I know you already mentioned that "it was failing", but can you also share the reason behind ?

gurkanindibay: You are right it does not fail until recently but started failing recently. Somehow it does not work
image

onurctirtir: I'm actually wondering the part "`Somehow it does not work`" because this shouldn't be the way we do the changes in this repo, can you please do some research about "why" ?

We should have a better reasoning than "tested & worked" approach.

onurctirtir: I'm actually wondering the part "`Somehow it does not work`", because this shouldn't be the way we do the changes in this repo, can you please do some research about "why" ?

We should have a better reasoning than "tested & worked" approach.

gurkanindibay: Actually I used the approach which is being used in checklist as above
$ git rev-parse tags/vX.Y.Z
<commitId>
$ git branch --contains <commitId>
release-X.Y
Then I found a stackoverflow post that I can handle it with just one command
https://stackoverflow.com/questions/15806448/git-how-to-find-out-on-which-branch-a-tag-is
image
However when it fails, I deep-dived and found out that the command to query tags is git tag and changed it to this one. 
As you see the documentation,

onurctirtir: I believe, using `git branch -a --contains tags/<tag>` should also work, as in the link you shared.

onurctirtir: I believe, using `git branch -a --contains tags/<tag>` should also work, as in the link you shared.

onurctirtir: I believe, using git branch -a --contains tags/<tag> should also work, as in the link you shared.

gurkanindibay: I've found out that there were another issue about removing * symbol from current branch so can not match the current branch

gurkanindibay: Now fixed without changing the command. As I said before there was an issue about parsing * sign.

onurctirtir (approved):
Now LGTM

PR #213, opened 2021-12-22
Use black to lint all python files

hanefi:
This PR contains all the changes after `black .` is ran. Also a new Github Actions Workflow is introduced to make sure we keep the formatting consistent.


onurctirtir:
see https://github.com/citusdata/tools/pull/291

PR #216, opened 2022-03-11
Adds beta support for releases

gurkanindibay:
1. Adds a new unit test branch for beta packages all-citus-unit-tests-beta to be able to test beta packages
2. Updates Unit test templates since Unit test branches updated to 10.2.4.
3. Adds -s flag for pytest in test pipelines to be able to see the logs in successful executions
4. Strips release stage (beta) from version to be able to get postgres version properly
5. Fixed centos-8 mirror problem for packaging tests (Fixed since tests are failing)

TODO:  After merging this task I need to move v0.8.12 tag into develop branch

### DIFF
--- a/packaging_automation/citus_package.py
+++ b/packaging_automation/citus_package.py
@@ -269,9 +269,10 @@ def build_packages(github_token: non_empty(non_blank(str)),
     os_name, os_version = decode_os_and_release(platform)
     release_versions, nightly_versions = get_postgres_versions(os_name, input_output_parameters.input_files_dir)
     signing_credentials = get_signing_credentials(signing_credentials.secret_key, signing_credentials.passphrase)
-
-    package_version = get_package_version_without_release_stage_from_pkgvars(input_output_parameters.input_files_dir)
-    write_postgres_versions_into_file(input_output_parameters.input_files_dir, package_version)
+    if platform != "pgxn":
+        package_version = get_package_version_without_release_stage_from_pkgvars(
+            input_output_parameters.input_files_dir)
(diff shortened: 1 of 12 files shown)


packaging_automation/citus_package.py
```diff
@@ -313,6 +315,15 @@ def get_package_version_from_pkgvars(input_files_dir: str):
     return package_version
 
 
+def get_package_version_without_release_stage_from_pkgvars(input_files_dir: str):
```
onurctirtir, line 318: What does release stage mean here ? Or, can you give an example about what does this function return for an example input, for me to better understand ?

test-images/debian-bullseye/Dockerfile
```diff
@@ -36,7 +36,6 @@ RUN apt-get update \
 RUN apt-get install -y postgresql-${PG_MAJOR}-citus-${CITUS_MAJOR_VERSION}=${CITUS_VERSION}.citus-${FANCY} \
                        postgresql-$PG_MAJOR-hll=${HLL_VERSION} \
                        postgresql-$PG_MAJOR-topn=${TOPN_VERSION} \
-    && apt-get purge -y --auto-remove curl \
```
onurctirtir, line 39: umm, why ?

gurkanindibay: We are using curl in the below block starting with line 55

packaging_automation/citus_package.py
onurctirtir: what about we call it as release_type then ?

PR #217, opened 2022-03-14
Adds beta release support for docker images

gurkanindibay:


packaging_automation/common_tool_methods.py
```diff
@@ -85,6 +86,18 @@ def get_minor_project_version(project_version: str) -> str:
     return f'{project_version_details["major"]}.{project_version_details["minor"]}'
 
 
+def does_project_version_contain_release_stage(project_version: str) -> bool:
```
onurctirtir, line 89: maybe:

```suggestion
def project_version_contains_release_type(project_version: str) -> bool:
```

packaging_automation/common_tool_methods.py
```diff
@@ -171,7 +184,12 @@ def is_merge_commit(commit: Commit):
 @validate_parameters
 def get_version_details(version: is_version(str)) -> Dict[str, str]:
     version_parts = version.split(".")
-    return {"major": version_parts[0], "minor": version_parts[1], "patch": version_parts[2]}
+    release_stage = 'stable'
+    if len(version_parts) == 3:
+        if '_' in version:
+            stage_parts = version_parts[2].split('_')
+            release_stage = stage_parts[1]
```
onurctirtir, line 191: Maybe put some comments and consider using release_type instead of release_stage:
Also, can it ever be the case that `len(version_parts)` is not equal to 3 ?

```suggestion
    release_type = 'stable'
    if '_' in version:
        # Having '_' in version string means that the last part of
        # it explicitly specifies release type.
        patch_version_parts = version_parts[2].split('_')
        release_type = patch_version_parts[1]
```

packaging_automation/tests/test_common_tool_methods.py
```diff
@@ -79,7 +79,7 @@ def test_remove_paranthesis_from_string():
 
 
 def test_get_version_details():
-    assert get_version_details("10.0.1") == {"major": "10", "minor": "0", "patch": "1"}
+    assert get_version_details("10.0.1") == {"major": "10", "minor": "0", "patch": "1","stage":"stable"}
```
onurctirtir, line 82: "stage" -> "release_type" ?

packaging_automation/common_tool_methods.py
gurkanindibay: Fixed as you suggested. Only used release_stage instead of release_type

packaging_automation/common_tool_methods.py
gurkanindibay: Used a method instead of the explicit line. This method name explains better I think

packaging_automation/tests/test_common_tool_methods.py
gurkanindibay: As I said stage is more specific than release_type

packaging_automation/common_tool_methods.py
```diff
@@ -171,7 +184,12 @@ def is_merge_commit(commit: Commit):
 @validate_parameters
 def get_version_details(version: is_version(str)) -> Dict[str, str]:
     version_parts = version.split(".")
-    return {"major": version_parts[0], "minor": version_parts[1], "patch": version_parts[2]}
+    release_stage = 'stable'
+    if len(version_parts) == 3:
```
onurctirtir, line 188: asking one more time, can it ever be different than 3 ?

onurctirtir (commented):
To me, release_stage sounds like something that corresponds to stages of a release, as in: preparation, deployment and testing etc. ...

For this reason, I'm more still inclined to using release_type or something similar in order to make it clear that it doesn't determine of which stage / step we are in the release, but type of the release.

packaging_automation/common_tool_methods.py
gurkanindibay: It is just an additional check I can remove if you want

gurkanindibay:
> To me, release_stage sounds like something that corresponds to stages of a release, as in: preparation, deployment and testing etc. ...
> 
> For this reason, I'm more still inclined to using release_type or something similar in order to make it clear that it doesn't determine of which stage / step we are in the release, but type of the release.

For usage of keyword stage I referenced wiki page which seems reasonable for me
https://en.wikipedia.org/wiki/Software_release_life_cycle#Stages_of_development

onurctirtir:
> > To me, release_stage sounds like something that corresponds to stages of a release, as in: preparation, deployment and testing etc. ...
> > For this reason, I'm more still inclined to using release_type or something similar in order to make it clear that it doesn't determine of which stage / step we are in the release, but type of the release.
> 
> For usage of keyword stage I referenced wiki page which seems reasonable for me https://en.wikipedia.org/wiki/Software_release_life_cycle#Stages_of_development

Then now that makes sense to me, thank you for the link :)

packaging_automation/common_tool_methods.py
onurctirtir: maybe make it an assert then

onurctirtir (approved):

packaging_automation/common_tool_methods.py
gurkanindibay: made it assert

PR #221, opened 2022-04-26
Adds minor version support for azure_gdpr  

gurkanindibay:
Azure gdpr project has minor version not patch version like hll so I added minor version support for this project

After merging this PR, I will move v0.8.13 tag to develop latest commit

### DIFF
--- a/packaging_automation/tests/test_citus_package.py
+++ b/packaging_automation/tests/test_citus_package.py
@@ -96,8 +96,8 @@ def test_build_packages():
             platform=PLATFORM)
         assert os.path.exists(postgres_version_file_path)
         config = dotenv_values(postgres_version_file_path)
-        assert config["release_versions"] == "13,14"
-        assert config["nightly_versions"] == "13,14"
+        assert config["release_versions"] == "12,13,14"
+        assert config["nightly_versions"] == "12,13,14"
 
 
 def test_get_required_package_count():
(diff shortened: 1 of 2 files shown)


packaging_automation/tests/test_citus_package.py
```diff
@@ -97,7 +97,7 @@ def test_build_packages():
         assert os.path.exists(postgres_version_file_path)
         config = dotenv_values(postgres_version_file_path)
         assert config["release_versions"] == "12,13,14"
-        assert config["nightly_versions"] == "12,13,14"
+        assert config["nightly_versions"] == "13,14"
```
gurkanindibay, line 100: Since nightly_versions for 11.x has changed, I needed to update nightly versions to fix the tests

onurctirtir: then should we drop 12 from L99 too ?

hanefi:
> After merging this PR, I will move v0.8.13 tag to develop latest commit

Why not create a new tag instead? I do not think that we should move tags, unless absolutely necessary.

packaging_automation/tests/test_citus_package.py
hanefi: Do we run the tests and assertions here? If so, I think we should see assertion failures on L99.

gurkanindibay: No tests are being run for 10.0.3 so release version is 12,13,14 but since nightlies are running on master and master is 11.x it fails.

hanefi:
> since azure_gdpr team is blocked I needed to put the tag into the branch

Please do not do this. If it is crucial to create a tag without merging PRs, you can instead consider using branch names to clone tools repo.

I still think you should create a new tag, instead of moving the current one.

hanefi:
> I still think you should create a new tag, instead of moving the current one.

@gurkanindibay decided to move the tag `v0.8.13` and performed a force-push instead of introducing a new tag as I recommended here.

PR #225, opened 2022-06-01
Adds latest spec template changes

gurkanindibay:
Rpm template has changed since NOTICE file  and columnar file additions. So templates and unit tests was needed to updated. 
Additionally, I updated the version to 10.2.4 to make it more up to date


packaging_automation/tests/files/citus.spec
```diff
@@ -81,6 +84,66 @@ echo %{pginstdir}/share/extension/%{sname}.control >> installation_files.list
 %doc %{pginstdir}/doc/extension/NOTICE-%{sname}
 
 %changelog
+* Tue Feb 01 2022 - Gurkan Indibay <gindibay@microsoft.com> 10.1.4.citus-1
+- Official 10.1.4 release of Citus
```
hanefi, line 88: Do we have missing entries for `10.2.4` here?

Is this not covered by the tests somehow?

rpmbuild may error out on this spec due to the mismatch between the `Version:` field and the first version in changelog entry. I am not 100% sure, but I recall observing some failures like this one in the past.

hanefi, line 88: Do we have missing entries for `10.2.4` here?

Is this not covered by the tests somehow?

rpmbuild may error out on this spec due to the mismatch between the `Version:` field and the first version in changelog entry. I am not 100% sure, but I recall observing some failures like this one in the past.

packaging_automation/tests/files/citus.spec
```diff
@@ -81,6 +84,66 @@ echo %{pginstdir}/share/extension/%{sname}.control >> installation_files.list
...
+
+* Fri Jul 09 2021 - Gurkan <gindibay@microsoft.com> 9.5.6.citus-1
+- Official 9.5.6 release of Citus
+
+* Thu Jul 08 2021 - Gurkan <gindibay@microsoft.com> 9.4.5.citus-1
+- Official 9.4.5 release of Citus
+
+* Thu Mar 18 2021 - Gurkan Indibay <gindibay@microsoft.com> 10.0.3.citus-1
+- Official 10.0.3 release of Citus
+
```
hanefi, line 146: Do we really need to have so many rows in the test file?

Does it not make sense to test the behaviour on a simpler and shorter spec?

packaging_automation/tests/files/citus.spec
```diff
@@ -7,12 +7,12 @@ Summary:	PostgreSQL-based distributed RDBMS
 Name:		%{sname}%{?pkginfix}_%{pgmajorversion}
 Provides:	%{sname}_%{pgmajorversion}
 Conflicts:	%{sname}_%{pgmajorversion}
-Version:	10.0.2.citus
+Version:	10.2.4.citus
```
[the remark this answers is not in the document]
gurkanindibay: Yes copy paste error. Actually it should be 10.1.4 which is former version of 10.2.4. We are using this file to assure that 10.2.4 is being updated correctly. Since we are changing the line, having a wrong version in citus.spec file does not affect the tests. We simply change the version line

packaging_automation/tests/files/citus.spec
gurkanindibay: Changed the version into 10.1.4 from 10.2.4. Changelog is correct. Wrong one is the version. This file is the initial version of the spec file before update_package_properties execution

packaging_automation/tests/files/citus.spec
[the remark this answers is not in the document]
gurkanindibay: Yes but I'm using actual files. I want to assure that only relevant parts are being changed. I had some cases where some regex'es replace irrevalant lines. Therefore I'm using actual file to make sure that only relevant changes are being made

hanefi (approved):

PR #236, opened 2022-08-08
Make changes to fix & as reference format

gledis69:
Currently,
```cpp
try{
	// ...
} catch (const runtime_error& err)
```
gets formatted to:
```cpp
try{
	// ...
} catch (const runtime_error&err)
```

This change fixes that.


gledis69:
Wrong way to change the config

PR #238, opened 2022-08-09
Adds exception support for postgres versions

gurkanindibay:
Currently, there is no mechanism to be able to  put an exception for ans os/release pair. With the pg_exclude.yml file, we can add exceptions for specific pg releases. If we set os/release pair as "all", all os/release pairs will be excluded

### DIFF
--- a/packaging_automation/citus_package.py
+++ b/packaging_automation/citus_package.py
@@ -285,6 +285,7 @@ def get_docker_image_name(platform: str):
 @validate_parameters
 # disabled since this is related to parameter_validations library methods
 # pylint: disable=no-value-for-parameter
+# pylint: disable= too-many-locals
 def build_packages(github_token: non_empty(non_blank(str)),
                    platform: non_empty(non_blank(str)),
                    build_type: BuildType, signing_credentials: SigningCredentials,
@@ -308,7 +309,6 @@ def build_packages(github_token: non_empty(non_blank(str)),
                                                         os_name] == PostgresVersionDockerImageType.single \
         else postgress_versions_to_process
 
-    print(f"Postgres Versions: {str_array_to_str(postgres_docker_extension_iterator)}")
     docker_image_name = get_docker_image_name(platform)
     output_sub_folder = get_release_package_folder_name(os_name, os_version)
     input_output_parameters.output_dir = f"{input_output_parameters.output_dir}/{output_sub_folder}"
--- a/packaging_automation/citus_package.py
+++ b/packaging_automation/citus_package.py
@@ -302,7 +302,7 @@ def build_packages(github_token: non_empty(non_blank(str)),
         raise ValueError("PACKAGING_PASSPHRASE should not be null or empty")
     postgress_versions_to_process = release_versions if build_type == BuildType.release else nightly_versions
 
-    postgres_docker_extension_iterator = "all" if platform_postgres_version_source[
+    postgres_docker_extension_iterator = ["all"] if platform_postgres_version_source[
                                                       os_name] == PostgresVersionDockerImageType.single \
         else postgress_versions_to_process
 
--- a/packaging_automation/citus_package.py
+++ b/packaging_automation/citus_package.py
@@ -20,8 +20,7 @@ from .common_tool_methods import (DEFAULT_ENCODING_FOR_FILE_HANDLING,
                                   get_supported_postgres_nightly_versions,
                                   get_supported_postgres_release_versions,
                                   platform_names,
-                                  run_with_output, str_array_to_str,
-                                  supported_platforms,
+                                  run_with_output, supported_platforms,
                                   transform_key_into_base64_str)
 from .packaging_warning_handler import validate_output
 
--- a/packaging_automation/citus_package.py
+++ b/packaging_automation/citus_package.py
@@ -162,14 +162,17 @@ def write_postgres_versions_into_file(input_files_dir: str, package_version: str
                                       platform: str = ""):
     # In ADO pipelines function without os_name and platform is used. If these parameters are unset
     if not os_name:
+        print("os name is empty")
         release_versions = get_supported_postgres_release_versions(f"{input_files_dir}/{POSTGRES_MATRIX_FILE_NAME}",
                                                                    package_version)
         nightly_versions = get_supported_postgres_nightly_versions(f"{input_files_dir}/{POSTGRES_MATRIX_FILE_NAME}")
     else:
+        print(f"os: {os_name} platform: {platform}")
         release_versions, nightly_versions = get_postgres_versions(os_name=os_name, platform=platform,
                                                                    input_files_dir=input_files_dir)
     release_version_str = ','.join(release_versions)
     nightly_version_str = ','.join(nightly_versions)
+    print(f"Release versions: {release_version_str}, Nightly versions: {nightly_version_str}")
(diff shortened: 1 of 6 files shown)


packaging_automation/citus_package.py
```diff
@@ -18,15 +20,15 @@
...
-                                  supported_platforms,
+                                  run_with_output, supported_platforms,
                                   transform_key_into_base64_str)
 from .packaging_warning_handler import validate_output
 
 GPG_KEY_NAME = "packaging@citusdata.com"
 
 POSTGRES_VERSION_FILE = "supported-postgres"
 POSTGRES_MATRIX_FILE_NAME = "postgres-matrix.yml"
+POSTGRES_EXCLUDE_FILE_NAME = "pg_exclude.yml"
```
hanefi, line 31: Now we have 3 files, and it is quite hard to differentiate their purposes. Can we move the contents of this file inside `postgres-matrix.yml` instead?

packaging_automation/citus_package.py
```diff
@@ -154,12 +156,21 @@ def get_signing_credentials(packaging_secret_key: str,
     return SigningCredentials(secret_key=secret_key, passphrase=passphrase)
 
 
-def write_postgres_versions_into_file(input_files_dir: str, package_version: str):
-    release_versions = get_supported_postgres_release_versions(f"{input_files_dir}/{POSTGRES_MATRIX_FILE_NAME}",
-                                                               package_version)
-    nightly_versions = get_supported_postgres_nightly_versions(f"{input_files_dir}/{POSTGRES_MATRIX_FILE_NAME}")
+def write_postgres_versions_into_file(input_files_dir: str, package_version: str, os_name: str = "",
+                                      platform: str = ""):
+    # In ADO pipelines function without os_name and platform is used. If these parameters are unset
```
hanefi, line 161: I do not really follow this comment. Why and how do we call this with missing parameters?

packaging_automation/citus_package.py
```diff
@@ -210,15 +221,22 @@ def sign_packages(sub_folder: str, signing_credentials: SigningCredentials,
...
-def get_postgres_versions(os_name: str, input_files_dir: str) -> Tuple[List[str], List[str]]:
-    if platform_postgres_version_source[os_name] == PostgresVersionDockerImageType.single:
-        release_versions = ["all"]
-        nightly_versions = ["all"]
-    else:
-        package_version = get_package_version_without_release_stage_from_pkgvars(input_files_dir)
-        release_versions = get_supported_postgres_release_versions(f"{input_files_dir}/{POSTGRES_MATRIX_FILE_NAME}",
-                                                                   package_version)
-        nightly_versions = get_supported_postgres_nightly_versions(f"{input_files_dir}/{POSTGRES_MATRIX_FILE_NAME}")
+def get_postgres_versions(platform: str, input_files_dir: str) -> Tuple[List[str], List[str]]:
```
hanefi, line 224: I have trouble reading and understanding this file. Can we have a simpler logic for set difference maybe?

packaging_automation/citus_package.py
```diff
@@ -264,31 +284,38 @@ def get_docker_image_name(platform: str):
 @validate_parameters
 # disabled since this is related to parameter_validations library methods
 # pylint: disable=no-value-for-parameter
+# pylint: disable= too-many-locals
```
hanefi, line 287: I think limiting the number of local variables to 15 is a good practice. How about splitting this into smaller functions and not adding an exception to pylint?

packaging_automation/citus_package.py
```diff
@@ -264,31 +284,38 @@ def get_docker_image_name(platform: str):
...
-        write_postgres_versions_into_file(input_output_parameters.input_files_dir, package_version)
+        write_postgres_versions_into_file(input_output_parameters.input_files_dir, package_version, os_name, platform)
 
     if not signing_credentials.passphrase:
         raise ValueError("PACKAGING_PASSPHRASE should not be null or empty")
+    postgress_versions_to_process = release_versions if build_type == BuildType.release else nightly_versions
+
+    postgres_docker_extension_iterator = ["all"] if platform_postgres_version_source[
+                                                        os_name] == PostgresVersionDockerImageType.single \
+        else postgress_versions_to_process
```
hanefi: This is a bit too hard to read. Can we rewrite the same logic?

```suggestion
    if platform_postgres_version_source[os_name] == PostgresVersionDockerImageType.single :
        postgres_docker_extension_iterator = ["all"] 
    else :
        postgres_docker_extension_iterator = postgress_versions_to_process
```

packaging_automation/citus_package.py
hanefi: One idea is to use set difference to remove the excluded items from supported items. I roughly have the following idea:

```py
>>> d1 = {1:1, 2:2, 3:3, "foo":4}
>>> d2 = {1:1, 2:4, 3:3, "bar":5, 6:6}
>>> dict(set(d1.items()).difference(set(d2.items())))
{'foo': 4, 2: 2}
```

packaging_automation/citus_package.py
gurkanindibay: Actually supported-postgres is just a auto-generated file from postgres-matrix.yml. So we have just 2 files for users.
Actually we will not need this file,pg_exclude.yml, quite common so I prefer not adding this option into postgres_matrix file. Additionally, postgres matrix includes all versions which is some sort of a citus postgres version support matrix file. This file will be very long and adding additional things will make the file more difficult to read.
However, it is not a bad idea. If it is needed, I need to open a PR about that and merge them in the future since there are more prioritized tasks waiting my attention in my work queue.

packaging_automation/citus_package.py
gurkanindibay: ADO pipelines uses this function without os_name and platform. There is no need for ADO pipelines since there is no nightly builds there. I just want to emphasized the need for default values for parameters

packaging_automation/citus_package.py
gurkanindibay: You're right it is very long statement to use one liner check

packaging_automation/citus_package.py
gurkanindibay: The method is 33 lines long which seems very reasonable to me. Splitting the method is not a good option which would hurt the readability I think. If it is required to fit into 15 local variable for here, I can use single tuple values for decode_os_and_release and get_postgres_versions. However, in that case I need to access values with numerical indexes which will make it hard to read

packaging_automation/citus_package.py
gurkanindibay:
```suggestion

postgres_versions = [12, 13, 14]
exclude_dict = {"ol/7": [14]}
platform = "ol/7"

if platform in exclude_dict:
    filtered_postgres_versions = set(postgres_versions).difference(set(exclude_dict[platform]))
    print(filtered_postgres_versions)
```

```suggestion
release_versions = [v for v in release_versions if v not in exclude_dict_release[platform_key_release]]
```

Second one seems more clear to me. Performance is not a top requirement for us I know but in the first one there are two cast operations.

Please check it again and choose whichever you want. It is OK for me

gurkanindibay:
@hanefi I addressed all of the review notes. Could you check it again? Thanks

hanefi (approved):
I do not like the current design, but I will approve this in the hopes that we fix it soon. The main problem is captured in https://github.com/citusdata/packaging/issues/958

PR #240, opened 2022-08-31
Adds enterprise changes into template files

gurkanindibay:


.github/workflows/package-cloud-download-schedule.yml
```diff
@@ -9,11 +9,7 @@ env:
   PACKAGE_CLOUD_ADMIN_API_TOKEN: ${{ secrets.PACKAGE_CLOUD_ADMIN_API_TOKEN }}
 on:
   schedule:
-    - cron: "0 16 * * *"
-
-  push:
-    branches:
-      - "**"
+    - cron: "0 16 * * *" # https://crontab.guru/#0_16_*_*_*
```
onurctirtir, line 12: thanks for adding the link

.github/workflows/package-cloud-download-schedule.yml
```diff
@@ -9,11 +9,7 @@ env:
   PACKAGE_CLOUD_ADMIN_API_TOKEN: ${{ secrets.PACKAGE_CLOUD_ADMIN_API_TOKEN }}
 on:
   schedule:
-    - cron: "0 16 * * *"
-
-  push:
-    branches:
-      - "**"
```
onurctirtir, line 16: why removing this ?

packaging_automation/tests/test_common_tool_methods.py
```diff
@@ -100,7 +100,7 @@ def test_replace_line_in_file():
                   errors=DEFAULT_UNICODE_ERROR_HANDLER) as reader:
             content = reader.read()
             lines = content.splitlines()
-            assert lines[5] == replace_str
+            assert lines[6] == replace_str
```
onurctirtir, line 103: couldn't follow why we're doing this

packaging_automation/templates/citus.spec.tmpl
```diff
@@ -35,7 +36,15 @@ commands.
...
-%configure PG_CONFIG=%{pginstdir}/bin/pg_config --with-extra-version="%{?conf_extra_version}" --with-security-flags
+
+currentgccver="$(gcc -dumpversion)"
+requiredgccver="4.8.2"
+if [ "$(printf '%s\n' "$requiredgccver" "$currentgccver" | sort -V | head -n1)" != "$requiredgccver" ]; then
+    echo ERROR: At least GCC version "$requiredgccver" is needed to build with security flags
+    exit 1
+fi
+
+%configure PG_CONFIG=%{pginstdir}/bin/pg_config --with-extra-version="%{?conf_extra_version}" --with-security-flags CC=$(command -v gcc) --without-pg-version-check
```
onurctirtir: why we're adding `--without-pg-version-check` ?

packaging_automation/templates/citus.spec.tmpl
```diff
@@ -2,6 +2,7 @@
 %global pgpackageversion 11
 %global pginstdir /usr/pgsql-%{pgpackageversion}
 %global sname citus
+%global debug_package %{nil}
```
onurctirtir, line 5: do we really need this ?

.github/workflows/package-cloud-download-schedule.yml
gurkanindibay: I do not want statistics packagecloud download pipelines to execute anymore since it takes too much time nd too much resource. I don't make any changes there anymore

packaging_automation/tests/test_common_tool_methods.py
gurkanindibay: since citus.spec file changed, tests need to be changed as well
%global debug_package %{nil} 
line has been added so Summary line shifted downwards

packaging_automation/templates/citus.spec.tmpl
gurkanindibay: good catch. I added the beta version config line by mistake. Removed it

packaging_automation/templates/citus.spec.tmpl
onurctirtir: ping @gurkanindibay

gurkanindibay: Yes without it we build fails. It is from enterprise config file 
You can check it here
https://github.com/citusdata/packaging/blob/0370f3337e160587d698358ab3c28f0271fde13c/citus-enterprise.spec#L6

gurkanindibay: and this one
https://github.com/citusdata/packaging/blob/bc964730234f14bb5e72600b219b2403f6adc87b/citus.spec#L5

gurkanindibay: Sorry I missed this one @onurctirtir

.github/workflows/package-cloud-download-schedule.yml
onurctirtir: It will still run as a cronjob run but will not get triggered on each branch push, right ? If that's is the case, I'm ok.

onurctirtir: But even if that's the case, that could be done in another PR since it's hard to relate it to enterprise changes.

gurkanindibay: Yes it's that case. Schedule is unaffected .Everybody is so busy so I wanted to handle and make it approved in this PR
I will be more careful for further PR's

onurctirtir (approved):

PR #250, opened 2022-10-04
Add option for custom shared libs

gledis69:


halilozanakgul (changes requested):
Why not also go ahead and create those extensions while we're at it?

PR #255, opened 2022-10-25
Fixes for problems that prevents automation on packaging 

gurkanindibay:
Problems fixed in this PR
* An unnecessary space in rpm template
* Tools repo is being added submodule when executing git add after Dockerfile updates.
* Postgres version change is being written into the changelog in all cases

### DIFF
--- a/packaging_automation/update_docker.py
+++ b/packaging_automation/update_docker.py
@@ -96,20 +96,17 @@ def create_directory_if_not_exists(dest_file_name):
         os.makedirs(dir_name)
 
 
-def get_new_changelog_entry(project_version: str, postgres_version: str = ""):
+def get_new_changelog_entry(project_version: str):
     header = f"### citus-docker v{project_version}.docker ({datetime.strftime(datetime.now(), '%B %d,%Y')}) ###\n"
     citus_bump_str = f"\n* Bump Citus version to {project_version}\n"
-    postgres_bump_str = f"\n* Bump PostgreSQL version to {postgres_version}\n"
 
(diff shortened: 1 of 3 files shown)


packaging_automation/update_docker.py
```diff
@@ -169,9 +163,14 @@ def read_postgres_versions(pkgvars_file: str) -> Tuple[str, str, str]:
     os.chdir(execution_path)
     pr_branch = f"release-{args.prj_ver}-{uuid.uuid4()}"
     run(f"git checkout -b {pr_branch}")
+
     update_all_docker_files(args.prj_ver, execution_path)
     run("git add .")
 
+    # git add command tries to add all the files, including cloned tools repo as submodule which is not needed
+    # this command removes the submodule from the commit
+    run("git rm -f --cached tools")
```
hanefi:
```suggestion
    run("git add --update .")
```

This command adds only those files that are already in the git worktree.

packaging_automation/update_docker.py
```diff
@@ -96,20 +96,17 @@ def create_directory_if_not_exists(dest_file_name):
         os.makedirs(dir_name)
 
 
-def get_new_changelog_entry(project_version: str, postgres_version: str = ""):
+def get_new_changelog_entry(project_version: str):
```
hanefi, line 99: Does this mean that you will manually add a new entry when we want to bump PG versions?

gurkanindibay: Yes because it is not an automated process ad we can update more than one version at once

hanefi (approved):

PR #257, opened 2022-10-27
Fix package tests pg15

gurkanindibay:


.github/workflows/package-tests.yml
```diff
@@ -52,7 +52,6 @@ jobs:
           - centos/8
           - centos/7
           - ol/7
-          - debian/stretch
```
gurkanindibay, line 55: Since debian is EOL, I dropped the support for it

onurctirtir (approved):

PR #261, opened 2022-11-14
Adds almalinux build image support into tools

gurkanindibay:
This PR is related to https://github.com/citusdata/packaging/pull/1006
To make almalinux work, I needed to change tools.


packaging_automation/citus_package.py
```diff
@@ -31,14 +31,19 @@
 POSTGRES_EXCLUDE_FILE_NAME = "pg_exclude.yml"
 
 docker_image_names = {
+    "almalinux": "almalinux",
+    "rockylinux": "almalinux",
     "debian": "debian",
+    "el/9": "almalinux-9",
```
gurkanindibay, line 37: Before this change, I was only using OS name to define docker image name. Sincei both centos and almalinux is being used as base image now, I added os version as well

packaging_automation/tests/test_citus_package.py
```diff
@@ -28,6 +28,10 @@
     "el/8": 3,
     "ol/7": 2,
     "ol/8": 3,
+    "almalinux/9": 3,
+    "rockylinux/9": 3,
+    "elnew/9": 3,
+    "olnew/9": 3,
```
onurctirtir: why do we have those "new" suffixes ?

packaging_automation/citus_package.py
```diff
@@ -279,6 +285,8 @@ def get_release_package_folder_name(os_name: str, os_version: str) -> str:
 
 
 def get_docker_image_name(platform: str):
+    if platform in docker_image_names.keys():
+        return docker_image_names[platform]
```
onurctirtir, line 293: Maybe comment:
```py
# Use hardcoded docker image names (e.g., almalinux-9) not the generic ones
# (e.g., almalinux) if docker_image_names tells us to do so.
```

gurkanindibay: Detailed the comments for the method

packaging_automation/tests/test_citus_package.py
gurkanindibay: it was my fault. Removed the new suffixes.

onurctirtir (approved):

PR #263, opened 2022-11-15
Adds support for debian/bookworm,ubuntu/kinetic

gurkanindibay:
Ubuntu/kinetic packagecloud support is not ready yet. After packagecloud support for ubuntu/kinetic, code should be added.

I could not add ubuntu/kinetic docker image either, since deb images depends on packagecloud and packagecloud does not have ubuntu/kinetic support yet.
However, still I want to keep the changes since I have meeting to add the support.


hanefi:
We could have the changes for debian/bookworm merged if we had separate PRs for the 2 distros. The changes look ok.

hanefi (approved):

PR #277, opened 2022-12-05
Updates requirements into latest version

gurkanindibay:
Dependabot alerts are enabled for this repo and I updated the requirements file with pip-compile command which is part of pip-tools. After updating the versions, some additional static code analysis results arised and fixed them as well.


hanefi:
Several questions:
- Where do we use this new file `packaging_automation/requirements_citus_pipeline.txt` ?
- How do you generate this file? Any documentation?
- The only difference between 2 files with names `packaging_automation/requirements*.txt` are the comments. Why do you think we need 2 generated files that differ only on comments?

gurkanindibay:
> Several questions:
> 
> * Where do we use this new file `packaging_automation/requirements_citus_pipeline.txt` ?
> * How do you generate this file? Any documentation?
> * The only difference between 2 files with names `packaging_automation/requirements*.txt` are the comments. Why do you think we need 2 generated files that differ only on comments?

1. In the new Citus pipelines
2. Just with pip-compile. It automatically updates with the latest versions
3. It was my bad. Now fixed it

hanefi:
I still do not understand why we need to generate and maintain 2 different requirements.txt files?

Why not we have a single file that works everywhere?

gurkanindibay:
> I still do not understand why we need to generate and maintain 2 different requirements.txt files?
> 
> Why not we have a single file that works everywhere?

@hanefi I removed the additional requirement file by installing python3.8 into all packaging images and setting python3.8 as default python3 for the python installed images. 
Before preparing this PR, I tried to install python3.8 from source (BTW python3.8 binaries does not exist in repositories for these OS/release pair), however I could not succeed and I see that I missed some prerequisite packages and now I fixed the problem.
Related PR is 
https://github.com/citusdata/packaging/pull/1020

Therefore I updated the PR header and content since the motive has changed with the new structure.

hanefi (approved):
Looks good now. However I suggest we wait until https://github.com/citusdata/packaging/pull/1020 is complete

PR #292, opened 2023-03-23
Adds pypi stats collector

gurkanindibay:


.github/workflows/pypi-statistics-schedule.yml
```diff
@@ -0,0 +1,32 @@
...
+  DB_PASSWORD: ${{ secrets.STATS_DB_PASSWORD }}
+  DB_HOST_AND_PORT: ${{ secrets.STATS_DB_HOST_AND_PORT }}
+  DB_NAME: ${{ secrets.STATS_DB_NAME }}
+on:
+  schedule:
+    - cron: "0 15 * * *"
+  push:
+    branches:
+      - "statistics/**"
+      - "dependabot/**"
```
hanefi, line 14: I do not understand why we run the builds on pushes to these branches.

packaging_automation/dbconfig.py
```diff
@@ -4,7 +4,7 @@
 from attr import dataclass
 from sqlalchemy import Column, INTEGER, TIMESTAMP, TEXT
 from sqlalchemy import create_engine
-from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import declarative_base
```
hanefi, line 7: The import order here looks weird. You can actually use isort to sort the imports properly.

For the time being, it is ok. However it would be better to introduce isort and flake8 support in a separate PR to improve the code readability a little bit. You can check out https://github.com/citusdata/citus/pull/6700 that sets a great example for a similar goal.

packaging_automation/tests/test_package_cloud_statistics_collector.py
```diff
@@ -68,7 +72,7 @@ def test_fetch_and_save_package_cloud_stats():
 
     records = session.query(PackageCloudDownloadStats).all()
 
-    assert len(records) == filtered_package_count * PACKAGE_SAVED_HISTORIC_RECORD_COUNT
+    assert len(records) > 0
```
hanefi, line 75: Is it enough to check for a positive number of records here? If we stop storing new statistics after a while, will this test fail, or will it silently ignore the problem?

packaging_automation/pypi_stats_collector.py
```diff
@@ -0,0 +1,81 @@
...
+        else:
+            existing_record_count += 1
+
+    session.commit()
+    print(
+        f"Process finished. New records: {new_record_count} Existing records: {existing_record_count}. Finished at {datetime.now()}"
+    )
+
+
+packages = ["django-multitenant"]
```
hanefi: IMO, it would be better to have the definition of this variable at the top of the file for better visibility. The readers would actually understand that we are only interested in `django-multitenant` today.

You may also want to use a better name to show your intent better.

packaging_automation/dbconfig.py
hanefi, line 7: Why did we need to do this change?

I am curious especially because we did not bump sqlalchemy version here.

.github/workflows/pypi-statistics-schedule.yml
gurkanindibay: I just want this workflow to trigger only in statistics development. This approach already exists in other statistics workflows

packaging_automation/dbconfig.py
gurkanindibay: Thanks I will use isort in the further PR's. There is already prospector which includes flake8

packaging_automation/pypi_stats_collector.py
gurkanindibay: Previously Fixed :)
Changed the order as you suggested. 
packages name is clear enough in this context since the stats to be collected is package
if you want it clearer I may change it to packages_to_download_stats but it seems overkill to me

packaging_automation/dbconfig.py
gurkanindibay: There was a warning as below
/home/runner/work/tools/tools/packaging_automation/dbconfig.py:34: MovedIn20Warning: The ``declarative_base()`` function is now available as sqlalchemy.orm.declarative_base(). (deprecated since: 2.0) (Background on SQLAlchemy 2.0 at: https://sqlalche.me/e/b8d9)

packaging_automation/tests/test_package_cloud_statistics_collector.py
gurkanindibay: We are dropping the table before inserting so this check assures record existence

hanefi:
A high level question: Why do we want this to be separate from the rest of the statistic collection scripts?

Why do we want to introduce a new schedule instead of updating `statistic-schedule`?

https://github.com/citusdata/tools/blob/ea2632e7aaccf7476439ec8a67031635167d24f5/.github/workflows/statistic-schedule.yml#L37-L40

Unrelated note: the name of the step I shared above is wrong. It has nothing to do with `docker pull`.

gurkanindibay:
Previous one took so much time to finish. Therefore, I don't want to add this one into existing one. In the future, we may seperate the existing ones as well. But right now there is no need

packaging_automation/pypi_stats_collector.py
hanefi: @gurkanindibay I do not like giving you a review where you silently ignore a portion of my comment. What is fixed here exactly? Did you actually rename the variable as I recommended above? If you decided to keep the name as it is, did you provide a simple explanation why?

It is not in my responsibility as a reviewer to control that you actually `fixed` something when you have a comment on a thread as little as `fixed`. IMO you need to put more effort into communication and making it easier for fellow colleagues to help you.

gurkanindibay: @hanefi Thanks for your proactive effort however I'm waiting for the unfinished actions to finish to complete my review notes. I see that you're trying to help me but I think it's better to wait to assign you to start your review. 

Let me finish all my comments and assign to you

hanefi: Also please be more careful when sharing `Fixed` as a comment on a thread.

I stop looking at this PR now. Please ping me on personal chat when you want me to take a look at this PR.

gurkanindibay: For the 'Fixed' comment, since it was the simple one I thought it is enough but I will give more detail as much as possible in the future.

gurkanindibay:
> A high level question: Why do we want this to be separate from the rest of the statistic collection scripts?
> 
> Why do we want to introduce a new schedule instead of updating `statistic-schedule`?
> 
> https://github.com/citusdata/tools/blob/ea2632e7aaccf7476439ec8a67031635167d24f5/.github/workflows/statistic-schedule.yml#L37-L40
> 
> Unrelated note: the name of the step I shared above is wrong. It has nothing to do with `docker pull`.

Changed the step name

hanefi (approved):

gurkanindibay:
I am grateful for the detailed feedback you provided, @hanefi. Thank you.

PR #308, opened 2023-04-24
Minor improvements

onderkalaci:
- Remove references to mx: As that is the current default
- Always add coordinator to the metadata: As that is the default behavior on managed service

### DIFF
--- a/citus_dev/citus_dev
+++ b/citus_dev/citus_dev
@@ -2,7 +2,7 @@
 """citus_dev
 
 Usage:
-  citus_dev make <name> [--size=<count>] [--port=<port>] [--use-ssl] [--no-extension] [--mx] [--destroy] [--init-with=<sql_file>] [--init-worker-with=<sql_file>] [--with-pgbouncer]
+  citus_dev make <name> [--size=<count>] [--port=<port>] [--use-ssl] [--no-extension] [--destroy] [--init-with=<sql_file>] [--init-worker-with=<sql_file>] [--with-pgbouncer]
   citus_dev restart <name> [--watch] [--port=<port>]
   citus_dev (start|stop) <name> [--port=<port>] [--force]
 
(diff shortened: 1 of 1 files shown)


citus_dev/citus_dev
```diff
@@ -65,9 +64,6 @@ def createNodeCommands(clustername, role, index=None, usessl=False, mx=False):
         run(f"openssl req -new -x509 -days 365 -nodes -text -out {clustername}/{nodename}/server.crt -keyout {clustername}/{nodename}/server.key -subj '/CN={nodename}'")
         run(f"chmod 0600 {clustername}/{nodename}/server.key")
 
-    if mx:
-        run(f"echo \"citus.replication_model = 'streaming'\" >> {clustername}/{nodename}/postgresql.conf")
-
```
thanodnl, line 70: Is this not a requirement anymore? Or has this setting be made the default now?

citus_dev/citus_dev
```diff
@@ -160,17 +154,12 @@ def main(arguments):
...
-            if size == 0:
-                run(f"psql -p {port} -c \"SELECT * from master_add_node('localhost', {port}, groupId => 0);\"")
-            else:
-                for i in range(size):
-                    workerport = port + 1 + i
-                    run(f"psql -p {port} -c \"SELECT * from master_add_node('localhost', {workerport});\"")
-
-                    if arguments["--mx"]:
-                        run(f"psql -p {port} -c \"SELECT start_metadata_sync_to_node('localhost', {workerport});\"")
+            # we always add the coodinator
```
thanodnl: I think this comment only makes sense in the context of mx. To reduce confusion I would _not_ have any comment here.

thanodnl (approved):
two small comments, otherwise this can be merged.

rajeshkt78:
@onderkalaci  Is it possible to add an option "--addWorker" to add a new worker to the existing cluster?
This will be very useful to try "rebalancing" and shard move operations..
If possible, even a "--removeWorker" also..

aykut-bozkurt:
> @onderkalaci Is it possible to add an option "--addWorker" to add a new worker to the existing cluster? This will be very useful to try "rebalancing" and shard move operations.. If possible, even a "--removeWorker" also..

Can you open an issue for that? Then, I can improve the tool by adding add/remove node operation.

citus_dev/citus_dev
onderkalaci: Yeah, that is the default for a while now

citus_dev/citus_dev
onderkalaci: makes sense

PR #339, opened 2023-09-21
Adds bookworm into test images

gurkanindibay:
Debian bookworm is added to package tests in this PR

### DIFF
--- a/test-images/debian-bookworm/Dockerfile
+++ b/test-images/debian-bookworm/Dockerfile
@@ -1,0 +1,88 @@
+FROM debian:bookworm
+ARG CITUS_VERSION
+# FOrmat should be XY and should not include dots e.g for 10.2.1=>102
+ARG CITUS_MAJOR_VERSION
+ARG PG_MAJOR
+ARG FANCY=1
+ARG HLL_VERSION=2.18.citus-1
+ARG TOPN_VERSION=2.6.0.citus-1
+
+ENV CITUS_VERSION ${CITUS_VERSION}
+
+ENV PG_MAJOR ${PG_MAJOR}
+
+ENV TZ=Europe/Istanbul
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+
+# install prequisities
+RUN apt-get update \
+    && apt-get install -y   lsb-release \
+                            apt-utils \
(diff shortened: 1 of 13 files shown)


test-images/debian-bookworm/Dockerfile
```diff
@@ -0,0 +1,88 @@
...
+
+
+
+
+
+
+
+
+
+
```
hanefi:
```suggestion
```
Why so many empty lines?

test-images/debian-bookworm/Dockerfile
```diff
@@ -0,0 +1,88 @@
...
+# install Citus
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+       ca-certificates \
+       curl \
+    && curl -s https://install.citusdata.com/community/deb.sh | bash
+RUN apt-get install -y postgresql-${PG_MAJOR}-citus-${CITUS_MAJOR_VERSION}=${CITUS_VERSION}.citus-${FANCY} \
+                       postgresql-$PG_MAJOR-hll=${HLL_VERSION} \
+                       postgresql-$PG_MAJOR-topn=${TOPN_VERSION} \
+    && rm -rf /var/lib/apt/lists/*12
```
hanefi, line 39: Why do we not merge these `RUN` commands?

gurkanindibay: in sake of debugging purposes. If there is an error in a command, I have issues to find which statement causes this error.
Since this is not an image to publish, merging them will not give a benefit to us

test-images/debian-bookworm/Dockerfile
gurkanindibay: there were empty spaces in other files and I formed this one from others. Removed empty spaces for all files now

test-images/debian-bookworm/Dockerfile
hanefi: There is actually a benefit: the resulting docker image will be smaller. Every apt command that does not include `rm -rf /var/lib/apt/lists/*` at the end will result in bigger docker images, and having to download bigger docker images result in longer CI run times.

hanefi (approved):
I do not like having continuous RUN commands in dockerfiles. However, I leave the decision to improve their performance to the author.

PR #352, opened 2024-02-23
Some random citus_dev improvements

JelteF:
- citus_dev: pass arguments directly to createNodeCommands
- Don't fsync by default
- Remove trailing whitespace
- Add editorconfig
- Add some basic gitignore stuff for citus_dev
- Add --no-lib flag to not load citus shared library
- Configure port in the config file instead of pg_ctl
- Configure logging the config file instead of pg_ctl
- remove nightlies for EOL distros

### DIFF
--- a/.github/workflows/build-citus-community-nightlies.yml
+++ b/.github/workflows/build-citus-community-nightlies.yml
@@ -32,7 +32,6 @@ jobs:
           - ubuntu/bionic
           - ubuntu/focal
           - ubuntu/jammy
-          - ubuntu/kinetic
 
     steps:
       - name: Checkout repository
--- a/citus_dev/citus_dev
+++ b/citus_dev/citus_dev
@@ -3,8 +3,8 @@
 
 Usage:
   citus_dev make <name> [--size=<count>] [--port=<port>] [--use-ssl] [--no-extension] [--no-lib] [--destroy] [--init-with=<sql_file>] [--init-worker-with=<sql_file>] [--with-pgbouncer] [--fsync]
-  citus_dev restart <name> [--watch] [--port=<port>]
-  citus_dev (start|stop) <name> [--port=<port>] [--force]
+  citus_dev restart <name> [--watch]
+  citus_dev (start|stop) <name> [--force]
 
 Options:
   --size=<count>           Number of workers to create when 0 the coordinator will be added as a worker [default: 2]
@@ -46,7 +46,7 @@ def run(command, *args, **kwargs):
     return result
 
 
-def createNodeCommands(clustername, role, arguments, index=None):
+def createNodeCommands(clustername, role, port, arguments, index=None):
     nodename = role
     if index != None:
         nodename += "%d" % index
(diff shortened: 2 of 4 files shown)


.github/workflows/build-citus-community-nightlies.yml
```diff
@@ -23,16 +23,13 @@ jobs:
...
-          - el/7
           - el/8
-          - ol/7
           - debian/buster
           - debian/bullseye
           - debian/bookworm
           - ubuntu/bionic
           - ubuntu/focal
           - ubuntu/jammy
-          - ubuntu/kinetic
```
thanodnl: @gurkanindibay could you double check these changes?

citus_dev/citus_dev
```diff
@@ -44,21 +46,33 @@ def run(command, *args, **kwargs):
...
+    if arguments['--fsync']:
+        run(f"initdb -D {clustername}/{nodename} --no-sync")
+    else:
+        run(f"initdb -D {clustername}/{nodename}")
 
-    shared_preload_libraries = ['citus', 'pg_stat_statements']
+    if arguments['--no-lib']:
+        shared_preload_libraries = ['pg_stat_statements']
+    else:
+        shared_preload_libraries = ['citus', 'pg_stat_statements']
```
thanodnl: This works for now, but it becomes more cumbersome if we need more preload libraries in the future.
```suggestion
    shared_preload_libraries = ['pg_stat_statements']
    
    # double negation, what does it mean?
    # we do add citus if we do _not_ provide the `--no-lib` flag
    # because citus needs to be at the beginning of the list we perform this check last, and insert `citus` at index 0, which is the beginning of the list
    if not arguments['--no-lib']:
        shared_preload_libraries.insert(0, 'citus')
```

citus_dev/citus_dev
```diff
@@ -188,7 +201,7 @@ def main(arguments):
         clustername = arguments["<name>"]
         port = int(arguments["--port"])
         for role in getRoles(clustername):
-            run(f'pg_ctl {pgctl_flags} start -D {clustername}/{role} -o "-p {port}" -l {role}_logfile')
+            run(f'pg_ctl {pgctl_flags} start -D {clustername}/{role} -o "-p {port}"')
```
thanodnl: can the port be removed if we have it in the config now?

possibly we would also need to remove the flag from start/restart, given it's a property of the cluster now, not a runtime argument.

citus_dev/citus_dev
```diff
@@ -2,7 +2,7 @@
 """citus_dev
 
 Usage:
-  citus_dev make <name> [--size=<count>] [--port=<port>] [--use-ssl] [--no-extension] [--destroy] [--init-with=<sql_file>] [--init-worker-with=<sql_file>] [--with-pgbouncer]
+  citus_dev make <name> [--size=<count>] [--port=<port>] [--use-ssl] [--no-extension] [--no-lib] [--destroy] [--init-with=<sql_file>] [--init-worker-with=<sql_file>] [--with-pgbouncer] [--fsync]
   citus_dev restart <name> [--watch] [--port=<port>]
   citus_dev (start|stop) <name> [--port=<port>] [--force]
```
thanodnl: Guess `--port` can be removed here

thanodnl (approved):
looking good, lets make sure before merging @gurkanindibay is ok with us removing old distributions. I don't know in what manner the tools repo is used for packaging/building/testing anymore.

.github/workflows/build-citus-community-nightlies.yml
gurkanindibay: We can remove ubuntu/kinetic however, we can not remove other ones. We still support el/7 and ol/7

PR #379, opened 2025-02-05
Add Dockerfiles to qa citus packages

m3hm3t:


packaging_qa/Debian.bullseye.dockerfile
```diff
@@ -0,0 +1,62 @@
...
+        postgresql-$PG_MAJOR-hll=2.18.citus-1 \
+        postgresql-$PG_MAJOR-topn=2.7.0.citus-1
+
+# Ensure PostgreSQL data directory exists and has correct permissions
+RUN mkdir -p /var/lib/postgresql/$PG_MAJOR/main && chown -R postgres:postgres /var/lib/postgresql && chmod -R 700 /var/lib/postgresql
+
+# Switch to postgres user before running PostgreSQL
+USER postgres
+
+# Initialize the database only if it's not already initialized
```
onurctirtir, line 38: why can we have the database already initialized?

m3hm3t: To make this more robust, we can check for `pg_control `in `/var/lib/postgresql/${PG_MAJOR}/main/global/`, which ensures that the database directory is actually valid before skipping initialization.

```
RUN if [ ! -f "/var/lib/postgresql/${PG_MAJOR}/main/global/pg_control" ]; then \
       echo "Initializing new PostgreSQL database..."; \
       /usr/lib/postgresql/${PG_MAJOR}/bin/initdb -D /var/lib/postgresql/${PG_MAJOR}/main; \
    else \
       echo "Existing PostgreSQL database detected, skipping initdb."; \
    fi
```

Let me know if you think this approach works, and I can push an update!

PR #385, opened 2026-01-25
Modify build matrix for community nightlies

ihalatci:
Updated the build matrix to include new platforms and remove deprecated ones.

### DIFF
--- a/.github/workflows/build-citus-community-nightlies.yml
+++ b/.github/workflows/build-citus-community-nightlies.yml
@@ -1,7 +1,7 @@
 name: Build and publish citus community nightly packages
 
 env:
-  MAIN_BRANCH: "all-citus"
+  MAIN_BRANCH: "ihalatci-update-all-citus"
   PACKAGING_PASSPHRASE: ${{ secrets.PACKAGING_PASSPHRASE }}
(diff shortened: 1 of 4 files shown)


.github/workflows/build-citus-community-nightlies.yml
```diff
@@ -1,7 +1,7 @@
 name: Build and publish citus community nightly packages
 
 env:
-  MAIN_BRANCH: "all-citus"
+  MAIN_BRANCH: "ihalatci-update-all-citus"
```
m3hm3t: documents also using this branch. Are we sure to update it?

If yes, it would be better to update documents also.

m3hm3t:
only fail in checks

https://github.com/citusdata/tools/actions/runs/21474952760/job/61877508392?pr=385

https://hub.docker.com/r/citus/packaging/tags

probably looking in repo above.

m3hm3t (approved):

.github/workflows/build-citus-community-nightlies.yml
ihalatci: This was a temperory change since they depend on each other. But let's try to push this through without this change

ihalatci:
Nightly workflow is failing as a block due to version config mismatch (from all-citus branch). It will be updated soon, PR open after this merge and new tag.

PR #387, opened 2026-02-05
Fix Docker templates: use explicit base image tags and update Alpine …

m3hm3t:
Fix Docker templates: use explicit base image tags and update Alpine build

- Add -bookworm suffix to Debian-based Dockerfile templates to ensure
  consistent base images (postgres-14, postgres-15, postgres-16,
  postgres-17, postgres-18, latest)
- Update Alpine template to use versioned LLVM packages (llvm19-dev,
  clang19, llvm19-libs) instead of generic clang/llvm
- Refactor Alpine build commands for better readability

Backport from citusdata/docker@7794206


ihalatci:
Closing as outdated. The Bookworm pinning was subsequently reverted in citusdata/docker, the PG14/15 generation targets have been removed, and the Alpine LLVM/Clang package changes no longer match the current PG18 Alpine base. Any remaining Alpine template synchronization should be handled in a new focused PR with updated tests.

PR #401, opened 2026-04-16
Use exported Citus requirements in test images

ihalatci:
## Summary
- replace the shared test-image Python dependency input with the exported requirements from the Citus failure-test environment
- move the test-image Python runtime from 3.8 to 3.12 so upstream mitmproxy 12.2.2 can be installed
- keep the image test runner aligned with the refreshed dependency set

## Context
This supports the Citus dependency refresh and the switch away from the forked mitmproxy reference in the test environment.

## Verification
- installed the shared test-image requirements in a clean Python 3.12 environment
- confirmed the resolved versions include mitmproxy 12.2.2, cryptography 46.0.7, pytest 9.0.3, Werkzeug 3.1.6, and tornado 6.5.5


ihalatci:
Superseded: these test-image dependency updates are being moved to the-process repo instead.

PR #411, opened 2026-06-29
diag: surface stdout+stderr on build_package failure

ihalatci:
Packaging build failures were masked by stderr-only docker pull preamble output. This now prints both stdout and stderr, including the return code, so the real container error is visible. No behavior change on success.

### DIFF
--- a/packaging_automation/citus_package.py
+++ b/packaging_automation/citus_package.py
@@ -367,7 +367,11 @@ def build_package(
     if output.stdout:
         print("Output:" + output.stdout)
     if output.returncode != 0:
-        raise ValueError(output.stderr)
+        raise ValueError(
+            f"Build failed (rc={output.returncode}).\n"
+            f"STDOUT:\n{output.stdout}\n"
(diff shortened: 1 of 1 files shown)


packaging_automation/citus_package.py
```diff
@@ -367,7 +367,11 @@ def build_package(
     if output.stdout:
```
alperkocatas: It looks like stdout will be printed twice on failure case. So, I recommend moving lines 367-368 after the error case. 

```suggestion
    if output.returncode != 0:
        raise ValueError(output.stderr)
        raise ValueError(
            f"Build failed (rc={output.returncode}).\n"
            f"STDOUT:\n{output.stdout}\n"
            f"STDERR:\n{output.stderr}"
        )
        
if output.stdout:
        print("Output:" + output.stdout)
```

alperkocatas (dismissed):
Dropped a nit comment.

serhatandic (approved):

PR #414, opened 2026-07-08
Add multi-arch (amd64+arm64) buildx support for alpine Docker image

ihalatci:
## What

Adds `linux/arm64` build capability to the Citus **alpine** Docker image by replacing the single-arch Python Docker SDK build/tag/push path in `packaging_automation/publish_docker.py` with `docker buildx build` subprocess calls.

Refs citusdata/citus#8612.

## Why

The base images (`postgres:18.4` / `postgres:18.4-alpine`) already publish `arm64/v8` manifests. The **alpine** Citus image builds Citus from source (`./configure && make install`) with **no arch-specific packages**, so it is arm64-capable today. The Debian images install pinned `.deb` packages that don't yet exist for arm64, so **only alpine** goes multi-arch here — every other image stays amd64-only with functionally unchanged behavior.

## Cross-repo dependencies

- **citusdata/packaging#1198** adds the gated arm64 community-nightly `.deb` build leg. It is coordinated with this PR and starts populating the arm64 package index required for future Debian-based Citus images, but it is **not a prerequisite for the alpine image** implemented here.
- A follow-up in **citusdata/docker** must add QEMU/buildx setup, set `DOCKER_BUILD_MULTI_ARCH`, and consume a new `citusdata/tools` release containing this change. Until that workflow change opts in, Docker publishing remains amd64-only.

## Changes

- **`get_platforms()`** — gates alpine multi-arch behind env var `DOCKER_BUILD_MULTI_ARCH` (truthy set `1/true/yes/on`). Default **OFF** → `linux/amd64` for all images, preserving today's behavior. When ON, only `alpine` → `linux/amd64,linux/arm64`.
- **`run_buildx_build()`** — one `docker buildx build` call per image:
  - publishing → `--push` (multi-arch manifests can't be `--load`ed; must go straight to the registry),
  - not-publishing + single-arch → `--load` (parity with old local-image behavior),
  - not-publishing + multi-arch → build-only (validates arm64 compile in PRs, output discarded).
- Rewrote the 3 build functions to make one buildx call each; the tagged flow passes the full semver cascade (e.g. `14`, `14.1`, `14.1.0`) as multiple `-t` flags in a single build+push.
- Removed dead SDK code (`import docker`, module-level clients, `flush_logs()`).

### Companion changes

- **Test:** relaxed `test_publish_tagged_docker_images_alpine` to drop the `images.get(...)` asserts and rely on `run_buildx_build`'s `check=True` (gate-agnostic; with the var ON alpine is build-only so there's no local image to `get`).
- **Workflow:** added `docker/setup-qemu-action@v3` + `docker/setup-buildx-action@v3` to `publish-docker-image-tests.yml`. **`DOCKER_BUILD_MULTI_ARCH` is intentionally left unset** so tools PR/unit CI stays amd64-fast; arm64 is exercised at publish time by the citusdata/docker workflows.

## Compatibility & follow-ups

- amd64-only images keep functionally-equivalent behavior (default-OFF gate).
- buildx `--push` requires a prior `docker login` — the citusdata/docker publish workflows already log in; buildx reads `~/.docker/config.json`, so ordering is compatible.
- **After merge:** a new `citusdata/tools` release tag must be cut and the citusdata/docker workflows repinned (currently `citusdata/tools@v0.8.36`).
- QEMU/buildx setup on the citusdata/docker **publish** workflows is handled separately.

## Risk

Emulated arm64 alpine source-build under QEMU is slow (30–60+ min). Mitigated by the default-OFF env gate — the tools repo never pays this cost unless a workflow explicitly opts in.

## Validation

Local: `py_compile` OK on both Python files; grep confirms zero leftover `docker_client`/`docker_api_client`/`flush_logs`/`import docker` refs; AST check confirms all test-imported symbols present and `flush_logs` removed. Full pytest (real builds cloning citusdata/docker) not run locally — it runs in CI.

Diff stat: 3 files changed, 53 insertions(+), 48 deletions(-).

### DIFF
--- a/packaging_automation/publish_docker.py
+++ b/packaging_automation/publish_docker.py
@@ -1,9 +1,9 @@
 import argparse
 import os
+import subprocess
 from enum import Enum
 from typing import Tuple, List
 
-import docker
 import pathlib2
 from parameters_validation import validate_parameters
 
@@ -91,9 +91,37 @@ docker_image_info_dict = {
 }
 DOCKER_IMAGE_NAME = "citusdata/citus"
 
-docker_client = docker.from_env()
+PLATFORM_AMD64 = "linux/amd64"
+PLATFORM_MULTI_ARCH = "linux/amd64,linux/arm64"
 
-docker_api_client = docker.APIClient()
+
+def get_platforms(docker_image_type: DockerImageType) -> str:
+    multi_arch_enabled = os.getenv("DOCKER_BUILD_MULTI_ARCH", "").lower() in (
+        "1",
+        "true",
+        "yes",
+        "on",
+    )
+    if docker_image_type == DockerImageType.alpine and multi_arch_enabled:
+        return PLATFORM_MULTI_ARCH
+    return PLATFORM_AMD64
+
+
+def run_buildx_build(
+    dockerfile: str,
+    platforms: str,
+    image_tags: List[str],
+    will_image_be_published: bool,
+):
+    command = ["docker", "buildx", "build", "--platform", platforms, "-f", dockerfile]
+    for image_tag in image_tags:
+        command.extend(["-t", image_tag])
+    if will_image_be_published:
+        command.append("--push")
+    elif platforms == PLATFORM_AMD64:
+        command.append("--load")
+    command.append(".")
+    subprocess.run(command, check=True)
 
 
 def regular_images_to_be_built(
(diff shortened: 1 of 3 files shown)


packaging_automation/tests/test_publish_docker.py
```diff
@@ -101,9 +101,6 @@ def test_publish_tagged_docker_images_alpine():
     try:
         run_with_output("git checkout -b docker-unit-test")
         publish_tagged_docker_images(DockerImageType.alpine, TAG_NAME, False)
```
KardelenCeren, line 103: In the future, maybe we could add tests here covering the new Buildx behavior, with the multi-arch Alpine path when `DOCKER_BUILD_MULTI_ARCH` is enabled.

packaging_automation/publish_docker.py
```diff
@@ -252,20 +258,16 @@ def publish_tagged_docker_images(
...
-        path=".",
+        platforms=get_platforms(DockerImageType.nightly),
+        image_tags=[docker_image_name],
+        will_image_be_published=will_image_be_published,
     )
-    flush_logs(logs)
     print("Nightly image build finished.")
 
     if will_image_be_published:
         print("Pushing nightly image...")
```
KardelenCeren, line 270: nit: Since `run_buildx_build()` performs the push via `--push`, we could remove the “Pushing nightly image...” log here. It may be misleading because it is printed after the build-and-push operation has already completed.

KardelenCeren (approved):

PR #426, opened 2026-09-01
Fix nightly PostgreSQL exclusion filtering

ihalatci:
## Summary
- filter nightly PostgreSQL exclusions from `nightly_versions` instead of `release_versions`
- update the directly affected nightly exclusion test expectation

## Why
The bug is latent while `exclude.nightly` is empty: adding any nightly exclusion currently replaces the nightly matrix with filtered `release_versions`. Filtering `nightly_versions` preserves the intended nightly matrix while removing only the excluded PostgreSQL versions.

This is required by the companion all-citus matrix change that excludes PostgreSQL 19: https://github.com/citusdata/packaging/pull/1215. Both changes are required, but they can land in either order.

### DIFF
--- a/packaging_automation/citus_package.py
+++ b/packaging_automation/citus_package.py
@@ -329,7 +329,7 @@ def get_postgres_versions(
     if exclude_dict_nightly and platform_key_nightly in exclude_dict_nightly:
         nightly_versions = [
             v
-            for v in release_versions
+            for v in nightly_versions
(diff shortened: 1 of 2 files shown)


ihalatci:
Superseded by #423, which merged the same nightly exclusion fix into develop and is included in the stable v0.8.39 tag. Closing rather than resolving the duplicate conflict.

